PR #9, opened 2015-10-22
update DiscardHand

DailyShana:
add Duel.CheckDiscardHand

updated
the return value of Duel.DiscardHand is boolean if reason contains REASON_COST.
Otherwise the return value is the number of operated cards.


DailyShana:
eh.. if we allow return value to be a number when reason do not contain REASON_COST,  EFFECT_DISCARD_HAND_CHANGE should only be applied to cost

PR #19, opened 2015-12-06
fix EVENT_BATTLE_DESTROYING/ED

DailyShana:
fix: If Dark Magician of Chao affected by Spellbook of Power battle destroyes and removes the attack target, the effect of Spellbook of Power cannot be activated. 
Because the destroyed attack target has reason REASON_EFFECT(removed by effect)+REASON_BATTLE+REASON_DESTROY

the modification of operations.cpp in https://github.com/Fluorohydride/ygopro/commit/fd5d5bf7272a6e4153a6b87d8610dccacb642bbf is enough.


salix5:
4e0295fab88326b216e54b6da644676b42a609c4
已修正
謝謝

DailyShana:
话说为什么不用 bitwise or

PR #22, opened 2015-12-14
SEGOC rules/Chain order

sidschingis:
Trigger Effects of Monsters in the Hand should also follow SEGOC rules, so that you can freely chose in which order to activate the effects.
It is currently not possible to build the following chain:
Normal Summon Star Seraph Scepter, while having Star Seraph Sovereignity in Hand
CL1 Star Seraph Sovereignity
CL2 Star Seraph Scepter


VanillaSalt:
http://yugioh-wiki.net/?%BC%EA%BB%A5%CD%B6%C8%AF

手札誘発は通常の誘発効果だけでチェーンを組み終えた後に発動できる。
通常の誘発効果より先に発動することはできないので注意が必要。
例：手札に《イリュージョン・スナッチ》が存在し、《氷帝メビウス》をアドバンス召喚した場合、どちらも任意の誘発効果だが、チェーン１に《イリュージョン・スナッチ》の手札誘発を発動する事はできない。

sidschingis:
I cannot seem to find any official sources of those rulings anymore.

The latest reference I could find was https://yugiohblog.konami.com/articles/?p=7099 which comes from the TCG, so it might be another TCG/OCG difference.

-Because both effects triggered at the same time, you decide what order they go on the Chain. Make Sovereignty’s effect Chain Link 1, and Scepter’s search Chain Link 2.

Is there any way to double check with Konami Japan (Contact Form etc.)?

salix5:
It seems that the order of optional tigger effetcs in hands is different in TCG.

PR #27, opened 2015-12-22
Fix illegal enum forward declaration of effect_flag and effect_flag2

IceYGO:
Specifying the size is mandatory. Fixes the compilation on GCC (debian).

### DIFF
--- a/effect.h
+++ b/effect.h
@@ -24,8 +24,8 @@ class effect;
 struct tevent;
 struct effect_set;
 struct effect_set_v;
-enum effect_flag;
-enum effect_flag2;
+enum effect_flag : uint64;
+enum effect_flag2 : uint64;
 
 class effect {
 public:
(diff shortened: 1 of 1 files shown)


DailyShana:
uint32 seems enough and
it is illegal in vs2010

IceYGO:
`uint32` is indeed enough as long as we do not add any other effect flag.
I used `uint64` just in case, but I can change it.

And yes, it is illegal with vs2010 since vs2010 does not support C++11 properly. This project already uses some C++11 features everywhere, I do not think this is a problem. Heck, vs2010 is 6 years old!

DailyShana:
effect::flag[2] is uint32 so uint64 is meaningless indeed.

salix5:
The enum is 90% full at this time.
How could uint32 enough even in 2015?

PR #35, opened 2016-01-16
select trigger

DailyShana:
Rough description:
1. List all cards to be activated
2. Let player select one of them or cancel(uncancelable for forced triggers)
   if select goto 3
   if cancel goto 4
3. the selected effect will be activated as chain N+1(N=number of selected triggers by both players), goto 1
4. add chains(and let the other player select triggers)

The reason is whether a trigger effect can be activated depends on the chain order due to chain limit(what this commit do) and cost(which is more complicated)


salix5:
想問一下
這個主要改的是什麼部分

salix5:
這個修改是要取消所有的PROCESSOR_SORT_CHAIN
改用其他方式處理嗎？

DailyShana:
PROCESSOR_SORT_CHAIN还用在QUICK_F类型的效果，目前此类效果都没有连锁限制

PR #36, opened 2016-02-06
add Debug.PreSummon

DailyShana:
https://github.com/Fluorohydride/ygopro/issues/1672


DailyShana:
It will be convenient to test cards like Shaddoll Fusion

PR #48, opened 2016-06-26
Correct Warning C4267 in Visual Studio

Zayelion:
Steps to reproduce:
- 1.) Download Source
- 2.) Add missing libs
- 3.) `premake4 vs2013`
- 4.) Compile `ocgcore` project for `x86` both `Debug` and `Release`

Expected result is a clean compile, result is [C4244](https://msdn.microsoft.com/en-us/library/th7a07tz.aspx) warnings in most the files.

I dont think this "adds" anything to the codebase but less warnings is always better I personally feel.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -1131,7 +1131,7 @@ void card::xyz_add(card* mat, card_set* des) {
 	mat->overlay_target = this;
 	mat->current.controler = PLAYER_NONE;
 	mat->current.location = LOCATION_OVERLAY;
-	mat->current.sequence = xyz_materials.size() - 1;
+	mat->current.sequence = (uint8)xyz_materials.size() - 1;
 	mat->current.reason = REASON_XYZ + REASON_MATERIAL;
 }
 void card::xyz_remove(card* mat) {
--- a/field.cpp
+++ b/field.cpp
@@ -113,7 +113,7 @@ void field::reload_field_info() {
 			if(pcard) {
 				pduel->write_buffer8(1);
 				pduel->write_buffer8(pcard->current.position);
-				pduel->write_buffer8(pcard->xyz_materials.size());
+				pduel->write_buffer8((uint8)pcard->xyz_materials.size());
 			} else {
 				pduel->write_buffer8(0);
 			}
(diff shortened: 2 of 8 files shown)


IceYGO:
:+1: Looks good to me.

Would be even better using the C++ `static_cast<>` instead.

Zayelion:
@salix5 take a look?

salix5:
I'm searching for data about C4244

DailyShana:
https://github.com/Fluorohydride/ygopro-core/blob/master/common.h#L12

Zayelion:
I removed the line in `common.h` and got a new but similar set of errors. Is there any type of value gained by ignoring this warning?

DailyShana:
>to ensure that the range of possible values will always be compatible with the types you are using.

All these integers will not exceed the max value of uint8 actually(bounded by the size of deck)

Zayelion:
That may happen to be the case now, but the ultimate check for it if any of the code is changed else where is turned off. Just seems unsafe to me. I feel while its could be seen as only stylistic it helps the overall longevity of the code base for newer developers to correct it.

DailyShana:
Complile warning just remind it is possible sth would happen but we know that sth will not happen so it is meaningless to consider it and closing that warning is safe.

IceYGO:
Disabling warnings is stupid.

Yes, maybe the current code is safe. But how can we be sure? What if we change something? What if we add more code?

DailyShana:
Explicit type conversion/cast do not reduce the risk of data loss, it just disable the warning of that line of the code.
It is no difference between disabling warning and explicit conversion.

It's clear that data loss cannot happen here because all size of these volumes is bounded by the size of deck and if it really happened, we should reconstruct the code, not just use type conversion/cast.

We don't need to worry about this and should focus on the logic of rule and card effects.

IceYGO:
Explicit type conversion will, as you said, explicitly says to the 
compiler "I know what I am doing here".
It is entirely different than disabling the warning everywhere, telling 
to the compiler "Do not check my entire code".

Again, if we add some code causing an error because of an implicit type 
conversion someday, we will be glad if the compiler tells us where it 
comes from. It's not just about this code, it's about good habits.

DailyShana:
The result of explicit type conversion and disable warning is the same.
"I know what I am doing here, so do not check my entire code"

IceYGO:
If everyone working on the project is absolutely certain that the entire 
code (current and future) will never have any problem caused by an 
implicit cast, then yes, you can keep this line.

Zayelion:
Im not certain, Im learning C++ by working on this project. The change I'm suggesting are just code clean up. I was taught as code becomes stable to clean up parts of it. Its also open source we dont know what type of people and of what skill level they will be joining the project.

The code changes I did are causing a higher level warning than what that line of code is suppressing. (It doesnt seem smart to suppress errors to me, it seems smart to fix them).

At the moment I cant do the major code changes I want to do so as I'm learning Im personally just focusing on making the flow of code less complex for people reading it. From looking at it myself and other professional I've asked the opinion of and comparing it to other C++ projects the overall code base is very much a mess.

Zayelion:
Is this gonna be merged or rejected? Letting it sit here to bit rot is not particularly polite.

salix5:
Sorry for the late reply.
Since I have very limited time now, I am afraid that I can only check changes that directly related to card or system bugs.

Zayelion:
... Then could someone else press the button?

DailyShana:
fixed in e6ac7e88b247a917a9427ebb9334a1921be98c22

PR #50, opened 2016-07-11
Added EFFECT_TRIPLE_TRIBUTE

edo9300:


DailyShana:
Monster with EFFECT_TRIPLE_TRIBUTE should not be used as 2 tributes.

PR #59, opened 2016-08-26
Update interpreter with string literal for sprintf

Hanmac:
adding "%s" as string literal does fix this warnings:
```warning: format not a string literal and no format arguments [-Wformat-security]```

this one is not fixed yet, i don't know if it might break something on some OS (probably 32bit?)
```warning: format ‘%u’ expects argument of type ‘unsigned int’, but argument 4 has type ‘std::__cxx11::list<std::pair<void*, unsigned int> >::size_type {aka long unsigned int}’ [-Wformat=]```


DailyShana:
fixed in 9e7c8ca88d6ad773f299d19bc1f01f1af496a6ab

PR #64, opened 2016-10-12
Add Card.SetEntityCode

Tianchenglipu:
```
c:SetEntityCode(code)
```

This will **completely** change "c" to another card named "code". That is, it's image, original code, text attack, text defense, text attribute, text race and setnames are changed to "code"'s.

Note that this function doesn't overwrite "c"'s effect to "code"'s. That is, if you change a "Blue-Eyes Alternative White Dragon" to a "Dark Eradicator Warlock", this card will still have effects of "Blue-Eyes Alternative White Dragon", not effects of "Dark Eradicator Warlock". To switch it's effect, you may have to:

```
c:SetEntityCode(29436665)
c:ResetEffect(38517737,RESET_CARD)
c29436665.initial_effect(c)
```

However, if you just wanted to switch "Blue-Eyes White Dragon"'s image to another, you will not have any problems.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -2932,3 +2932,10 @@ int32 card::is_can_be_xyz_material(card* scard) {
 			return FALSE;
 	return TRUE;
 }
+uint32 card::set_entity_code(uint32 entity_code) {
+	card_data dat;
+	::read_card(entity_code, &dat);
+	uint32 code = data.code;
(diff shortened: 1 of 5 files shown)


DailyShana:
Sorry, we don't consider effects that is unrelated to ocg here.

PR #66, opened 2016-11-17
add Card.GetFusionType and Card.GetFusionRace

nekrozar:
Add `Card.GetFusionType` and `Card.GetFusionRace`
Add `EFFECT_MUST_BE_FUSION_SUBSTITUTE`(use _Parasite Fusioner_)
(Maybe `EFFECT_CHANGE_FUSION_TYPE` and `EFFECT_CHANGE_FUSION_RACE` don't need.)


DailyShana:
I think the effect of _Parasite Fusioner_ should be independent of fusion attribute, fusion set code, etc.
This direction is not appropriate.

nekrozar:
However I can't think of any good ideas...

TheOnePharaoh:
EFFECT_CHANGE_FUSION_ATTRIBUTE is needed in Predaplant Sundew Kingii

PR #73, opened 2017-02-10
Add an option to make the Player 2 go first

IceYGO:
This is useful in some situations (AI duel, ...) and does not require a lot of changes (3 lines).

### DIFF
--- a/duel.h
+++ b/duel.h
@@ -84,4 +84,5 @@ private:
 #define DUEL_PSEUDO_SHUFFLE		0x10
 #define DUEL_TAG_MODE			0x20
 #define DUEL_SIMPLE_AI			0x40
+#define DUEL_PLAYER_2_FIRST		0x80
 #endif /* DUEL_H_ */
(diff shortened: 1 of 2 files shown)


DailyShana:
I think specifying who goes first should be excuted in the server code.
ocgcore only consider the game process.

IceYGO:
I think who goes first is definitely a part of the game process.
Implementing it in the server requires much more code.

On gframe, it is implemented by swapping the players at the beginning of the duel, but it really feels dirty:
https://github.com/Fluorohydride/ygopro/blob/master/gframe/tag_duel.cpp#L324

lllyasviel:
Then maybe there will be a DUEL_PLAYER_2_FIRST_TAG for symmetry.And then there will be more “#define”. If so, we may find it better to let these handled by server code instead of ocgcore.

IceYGO:
Tag duels do not require such change: they are already handled by the server code properly and do not require swapping players.

PR #74, opened 2017-02-10
add Duel.GetTossedCoinCount and Duel.GetTossedDiceCount

nekrozar:
Deal with the _Dueltainment_(https://github.com/Fluorohydride/ygopro-pre-script/blob/master/scripts/20AP-JP/c100213060.lua).

### DIFF
--- a/operations.cpp
+++ b/operations.cpp
@@ -3623,7 +3623,7 @@ int32 field::send_to(uint16 step, group * targets, effect * reason_effect, uint3
 						retgrave.insert(pcard);
 						raise_single_event(pcard, 0, EVENT_RETURN_TO_GRAVE, pcard->current.reason_effect, pcard->current.reason, pcard->current.reason_player, 0, 0);
 					} else {
-						tograve.insert(pcard);					
+						tograve.insert(pcard);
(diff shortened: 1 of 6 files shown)


operations.cpp
```diff
@@ -5051,6 +5051,7 @@ int32 field::toss_coin(uint16 step, effect * reason_effect, uint8 reason_player,
 			pduel->write_buffer8(count);
 			for(int32 i = 0; i < count; ++i) {
 				core.coin_result[i] = pduel->get_next_integer(0, 1);
+				core.tossed_dice_count[playerid]++;
```
mercury233: i guess it should be `tossed_coin_count`

DailyShana:
I think `EVENT_TOSS_DICE` and `EVENT_TOSS_COIN` are enough.

nekrozar:
@DailyShana I tried `EVENT_TOSS_DICE` and `EVENT_TOSS_COIN`, but _Barrel Dragon_ get error message.

DailyShana:
Since they are not used, I'll revert this.
and tossed count can be realized by global effects via `EVENT_TOSS_DICE` and `EVENT_TOSS_COIN`.

PR #76, opened 2017-02-17
Added Duel.IsDuelType()

edo9300:
I made this function a while ago, and i think now it would be needed, in case of adding the actual duel rules and field as an obsolete ruling and making backward compatible scripts


lllyasviel:
If it is for master rule 4.0, I do not think it is necessary because we can open a port with old ocgcore and rule 3.0. If there is two rules in one ocgcore, the code will be hard to maintain.

edo9300:
So if a sort of backward compatibly will be done, the old version will use the old ocgcore?

PR #83, opened 2017-04-03
new function get_disabled_field for link-related lua

ZisIsNotZis:
Unless all link/position related functions will always be written in C++, I find it necessary to have this function in lua.


DailyShana:
link summon proc is completed and do not need this

PR #87, opened 2017-04-07
Added GroupSelectUnselect()

edo9300:
This functions accepts 2 groups as inputs, one from where you can select a card, and the other from where you can unselect a card, and it outputs a single card. It can be used in controllated loops, to perform cancelable selections with a recursive check, it can be used with cards like tune magician, for a full support, to improve the pendulum summon and remove all the various checkboxed in case of some summoning limitations or even with an impoved synchro procedure, that will allow to cancel the summon even after the tuner selection. This is the client related pull request https://github.com/Fluorohydride/ygopro/pull/1963


edo9300:
@VanillaSalt if you want to try how this function can be used, this is an upgrade to the actual link summon, using this function, that allows the entire summon to be cancelable, without any use of the message boxes.
[link summon with selectunselect.txt](https://github.com/Fluorohydride/ygopro-core/files/944712/link.summon.with.selectunselect.txt)

PR #93, opened 2017-04-29
movetofield and Debug.AddCard() change

edo9300:
MoveToField has been modified to use the constant LOCATION_PZONE when moving a card to the Pendulum Zone, allowing a clearer distinction between effects that move a Pendulum monster to a Pendulum Zone and those that move it to the S&T Zone (ex. Equip effects) and making it more consistent with the previous changes caused by the new Master Rule.
Same has been done with the Debug.AddCard() function, now it will accept LOCATION_PZONE as parameter

### DIFF
--- a/field.h
+++ b/field.h
@@ -538,7 +538,7 @@ public:
 	int32 send_to(uint16 step, group* targets, card* target);
 	int32 send_to(uint16 step, group* targets, effect* reason_effect, uint32 reason, uint8 reason_player);
 	int32 discard_deck(uint16 step, uint8 playerid, uint8 count, uint32 reason);
-	int32 move_to_field(uint16 step, card* target, uint32 enable, uint32 ret, uint32 is_equip, uint32 zone);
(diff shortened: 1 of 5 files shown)


purerosefallen:
also i think a way to add cards to exmzone will be needed

edo9300:
@VanillaSalt @mercury233 Update, i added the support for a flag with the allowed sequences to move a card too, this will make Varrel Lord Dragon fully functional, without using separate functions https://github.com/Fluorohydride/ygopro-pre-script/pull/123, also now there's a sort of retro compatibility, for the locations, if there is no flag, then the sequences will be "assigned" as they were before, otherwise the flag will take the precedence

mercury233:
@edo9300 Sorry I still can't understand the code of ocgcore very well, so I can't review this.

@VanillaSalt Can you took a look at this and https://github.com/Fluorohydride/ygopro-core/pull/94 and https://github.com/Fluorohydride/ygopro-core/pull/92 ? The scripts of recent cards need those urgently.

PR #98, opened 2017-06-25
add Card.IsCanBeLinkMaterial, Card.GetLinkType and Card.IsLinkType

nekrozar:
https://twitter.com/YuGiOh_OCG_INFO/status/878900574036643840
Deal with the _Mecha Phantom Beast Liten_.


edo9300:
Wouldn't be better removing the monster type check? We know what happened with fusions and synchros with the last cards, that can use non monsters for a synchro and fusion summon (probably we'll get a similar card for xyzs). So in my opinion it would be better to "prepare" this functions in prevision of this kind of cards.

nekrozar:
@VanillaSalt @DailyShana @mercury233 What do you think of @edo9300 idea?

mercury233:
I think it should be ok to add `card::get_xyz_type` and `card::get_link_type` too.

nekrozar:
https://github.com/Fluorohydride/ygopro-core/pull/98/commits/0f4d4baef4fc26bc617d98198533309b161004d1 updated.
I'll add `card::get_xyz_type` later.

PR #103, opened 2017-07-27
fix Stargrail Swordsman Aurum 2nd effect

Lyris12:
add optional 4th parameter to duel_check_location for whether the card at sequence will leave the field during effect activation

Part 1 of my solution to https://github.com/Fluorohydride/ygopro/issues/2003 ; view Part 2 at https://github.com/Fluorohydride/ygopro-scripts/pull/904

Please let me know if you find any bugs in this code!

### DIFF
--- a/libduel.cpp
+++ b/libduel.cpp
@@ -1649,7 +1649,19 @@ int32 scriptlib::duel_check_location(lua_State *L) {
 	if(playerid != 0 && playerid != 1)
 		return 0;
 	duel* pduel = interpreter::get_duel_info(L);
-	lua_pushboolean(L, pduel->game_field->is_location_useable(playerid, location, sequence));
+	uint32 isOpen = pduel->game_field->is_location_useable(playerid, location, sequence);
+	uint32 hasleave = FALSE;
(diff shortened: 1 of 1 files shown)


mercury233:
I think we should add a parameter in `field::is_location_useable` to ignore the used location, instead of actually removing the card.

mercury233:
https://github.com/Fluorohydride/ygopro-core/pull/104 
I tried to make some changes too. Do you think mine will be better?

Lyris12:
Okay, and I'm curious as to the extra `0` parameter after `zone` on line 3048 of `card.cpp`

mercury233:
> int32 get_useable_count(card* pcard, uint8 playerid, uint8 location, uint8 uplayer, uint32 reason, uint32 zone = 0xff, __uint32* list = 0__, uint8 neglect_used = 0);

Lyris12:
Ah, okay! Thanks, and that works.

mercury233:
Thank you for your idea!

Lyris12:
You're welcome!

PR #113, opened 2017-08-28
Add Card.FilterEffect & Duel.FilterPlayerEffect

purerosefallen:
... Card.FilterEffect(card c, int code)
will return all the effects card c affected, which would be easy to check the effects like EFFECT_CANNOT_BE_XYZ_MATERIAL or sth else.

Duel.FilterPlayerEffect(int player, int code)
similar to Duel.IsPlayerAffectedByEffect, but also returns all the effects of a player


edo9300:
Wow, appaerently you had my same idea ;), even if my gunction is alled GetPlayer/CardEffect

purerosefallen:
This one is actually from KoishiPRO

Zayelion:
What is KoishiPRO?

edo9300:
Probably his client, ~~like edopro for me~~

purerosefallen:
The client used in 222 and YGOCC

mercury233:
So any example in card script?

mercury233:
https://github.com/Fluorohydride/ygopro-core/commit/c4e0a851dd8433c2aa165ba0d5666aa99aa4b93e

PR #114, opened 2017-09-10
Added cancelable attack back

edo9300:


DailyShana:
the attack cost should be executed before selecting attack target
>Ｑ：このカードの攻撃宣言を行うためにカードを手札に戻す行為と攻撃対象を選択する行為はどちらが先に行われますか？
　　（《ビッグバン・シュート》との兼ね合い。）
Ａ：カードを手札に戻してから攻撃対象を選択し攻撃宣言を行います。(10/11/18)
　　よって相手モンスターに装備された《ビッグバン・シュート》を手札に戻して攻撃宣言を行う場合装備されていたモンスターは攻撃対象に選択できません。(11/07/06)
http://yugioh-wiki.net/index.php?%A1%D4%B2%E2%A4%CE%C3%AB%A4%CE%A5%D5%A5%A1%A5%EB%A5%B3%A5%F3%A1%D5#faq

It says that if the attack announcement of _Mist Valley Falcon_ make _Big Bang Shot_ that is equiping to the opponent's monster return to hand, you cannot select that monster removed by the effect of Big Bang Shot as attack target.

edo9300:
@DailyShana as i can understand from all the various rulings, the field should be updated immediately after the cost is paid, since for ecample the ruling with snatch steal mark of the rose for mist valley falcon:

> Q. 元々の持ち主が相手の「霞の谷のファルコン」に、自分が発動した「薔薇の刻印」が装備されています。その「霞の谷のファルコン」のコントロールを得て自分が攻撃宣言をする際に、装備されている「薔薇の刻印」を手札に戻す事はできますか？

> A. 質問の状況のように、自身に装備されている「薔薇の刻印」を手札に戻し、「霞の谷のファルコン」で攻撃宣言する事はできます。ですが、「薔薇の刻印」を手札に戻した時点で「霞の谷のファルコン」のコントロールが元の相手に戻っているため、「霞の谷のファルコン」の攻撃対象を選択する事ができず、結果的に攻撃宣言は行われません。
translated is

> Q. The original owner is equipped with the opponent's "Falcon of the Kasumi Valley", "Roses of the Roses" that I invoked. Can you return the equipped "rose stamp" to your hand when you declare an attack by getting control of that "Falcon of Kasumi Valley"?

> A. As in the situation of the question, you can return the "Roses of Roses" equipped to them to your hand and declare an attack with "Falcon of Kasumi Valley". However, since the control of "Falcon of Kasumi Valley" has returned to the original opponent when returning "stamp of rose" to the hand, You can not select the attack target of "Falcon of Kasumi Valley", as a result no attack declaration will be done.

nb, i translated the answer by splitting it, infact if you translate all togheter, it translated in "you can" but if you translate from the comma it translates as "you cannot" and i think it's the correct translation.

this means, that immediately after the cost is paid, the field is updated, and in the case of big bang shot, the monster is no longer on the field, and in this casem the control has switched, meaning that you won't be able to select a target

edo9300:
@DailyShana ok, i put the cost where it was before, now the only thing left to do is to solve the problem of the instant field update

edo9300:
done

shadowfox87:
In both cases, it is still the same thing. The cost is paid once the attack target is selected->Big Bang Shot or Snatch Steal returns to the hand->The opponent's monster is banished (Big Bang Shot) or the control is switched (Snatch Steal) -> there is no longer an attack target because it is not on the opponent's field anymore->It becomes an attack replay since target is gone->Player can either directly attack or choose another attack target (they cannot cancel the attack since the cost has been paid).

The above ruling for Big Bang Shot does not mean that the cost was paid before selecting the attack target. The attack target was selected but the target was banished since big bang shot was removed from the field. Hence, the player has lost the target.

We even tested this scenario using the official online simulator, Legacy of the Duelist:
falconshot
In this case, big bang shot was returned to the hand AFTER the attack target was selected. Then, the monster was banished. Mist Valley Falcon's attack still continues and the player can select another appropriate attack target (attack replay). If there are no available attack targets, then it can attack directly.

Let us add a face-down trap card, Mirror Force to the opponent's field. In this case, the opponent can only activate mirror force AFTER the attack target has been selected. Attack declaration is complete once the target has been selected. This means that the Player has entered the Battle Step and gives the opponent a chance to respond.

mercury233:
_too busy to check it_

PR #116, opened 2017-09-18
add Card.GetColumnZone

nekrozar:
Deal with the _Mekk-Knight_.


edo9300:
@nekrozar wouldn't it better to make a Duel. function with the column number as parameter?

PR #120, opened 2017-10-04
Added more Link Related functions and effects

youruriel:
Added further Link Rating and Link Marker related functions and effects you can now increase,decrease and change the Link Rating, add,remove or completely Change the Link Marker Pointer, also you can now get or obtain the information of a card, if it would have or what Link Rating and Link Marker it had originally or previously on field.


DailyShana:
we do not consider effects that do not exist in ocg

mercury233:
but those effects may appear in ocg, so this pr should be merged

DailyShana:
Link number is the number of link markers, so it is totally different to level and rank. We should wait for those effect appears in ocg.

mercury233:
...Isn't level and rank the number of stars? However we do should wait until those effect appear.

DailyShana:
Link markers also play role in link status, not only count the link numbers.

purerosefallen:
I also agree to merge this, for it may appear in the future. (For now it supports some custom cards)

youruriel:
He guys thank you for the replies, unfortumately im in Hospital again and cant clearly  See via mobile  yet (its  horrible, yeah i know i Sound rest), für me and some of you  might  Experten such Effects  in near Future  that True and they  will appear.  But Shane is also right  we  should  wait til  more  from  official side  will be confirmed,  so I just  let the Codes here  pr  on my Repo  and you Decide  Individualiy, Who want to merke with.

mercury233:
It is suggested to add only 1 new feature in a commit or pull request.

youruriel:
Oh Sorry my issue, i just wanted to merge the forked master branch with a new branch :(. @mercury233

mercury233:
please use branch to open some new pull requests, thank you

PR #121, opened 2017-10-07
Added IsLevel/Rank/Link

edo9300:
Those are basic functions, and it's very weird they had not yet been added....


DailyShana:
They are redundant because there are GetLevel, GetRank and GetLink.

edo9300:
Well, wouldn't is better to write `Duel.GetMatchingGroup(Card.IsLevel,tp,loc,loc,nil,lv)` instead of using the aux function and write `Duel.GetMatchingGroup( aux.FilterEqualFunction(Card.GetLevel,lv),tp,loc,loc,nil)`?

DailyShana:
It think that's enough for usage.

mercury233:
Easier is better

PR #124, opened 2017-10-19
added EFFECT_FORCE_MZONE

edo9300:
This effect forces the player to use the zones put as value for a summon/set. This will require to change cards that use `Duel.GetLocationCount` for reasons different from control change and movetofield, like Ground Collapse to use `Duel.GetLocationCount(tp,LOCATION_MZONE,nil,LOCATION_REASON_COUNT)`


edo9300:
There still kaiju like cards that have to be updated, because their actual checks will ignore this effect, since they check for negative counts too

PR #125, opened 2017-10-22
Changed master rule options handling

edo9300:
details here https://github.com/Fluorohydride/ygopro/pull/2047


mercury233:
I think the two users who choose the rule should decide this, instead of ygopro.

PR #126, opened 2017-10-26
fix for duel win

AndreOliveiraMendes:
if two duel win are made at same time, only the first duel win takes effect, example:
relay soul and exodia: both are executed but relay soul is done first, so the game simply understand relay soul and disregard exodia, giving a loss by relay soul effect when it would be a draw.


purerosefallen:
repo gone, and the merge can't be done, and it's too outdated

PR #128, opened 2017-11-27
add EFFECT_LIMIT_MZONE_EXTRA

nekrozar:
Deal with the _Steelswarm Origin_.


edo9300:
@nekrozar I see a bit pointless making an ad hoc effect to handle this effect... It would be better to add some extra parameters to the existing one, otherwise we'll end up having too many effects to deal with

nekrozar:
I can't solve some problem in Mater Rule 3.
So this commit closes.

@edo9300 
Your idea is good.
However this commit closes because some problem happened.

PR #136, opened 2018-01-28
add metamethod to get the length of Group

purerosefallen:
so now we could use #g instead of g:GetCount() to simplify the scripts


DailyShana:
The style of script API has been decided at the beginning of the project years ago, I don't think we should change it.

mercury233:
It won't change the style too much, and the old style can still be used.

PR #141, opened 2018-03-17
Fixed various memory curruption instances

edo9300:
In libduel, when calling the special summon and change position
functions, the created card_sets were destroyed after the function block
ended, making the lua api crash (this error appears at random in some
builds, i still don't understand why sometimes all works fine and other
times this happens making the client/server crash). In operations
instead there was an out of bound access in the array, leading to the
corruption of other values.

### DIFF
--- a/field.h
+++ b/field.h
@@ -225,6 +225,7 @@ struct processor {
 	card_set discarded_set;
 	card_set destroy_canceled;
 	card_set delayed_enable_set;
+	card_set temp_set;
 	effect_set_v disfield_effects;
 	effect_set_v extram_effects;
 	effect_set_v extras_effects;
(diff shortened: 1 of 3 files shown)


DailyShana:
So, merging this without any analysis of the crash?

DailyShana:
Please provide the detail environment and steps that can reproduce the crash definitely.

DailyShana:
I mean, you should give the step so that other people can reproduce it.

edo9300:
Just activate any card that or uses Duel.SpecialSummon or Duel.ChangePosition

DailyShana:
I mean let other people reproduce the program which has this issue.

PR #145, opened 2018-04-09
fix check of EFFECT_EXTRA_RELEASE_NONSUM

mercury233:
Fix: _Lair of Darkness_ can make cards like _Destiny HERO - Plasma_ release cards from opponent because the val check of `EFFECT_EXTRA_RELEASE_NONSUM` didn't work in `Duel.GetReleaseGroup`.

We should update the script of _Lair of Darkness_ too.
```lua
function c59160188.relval(e,re,r,rp)
	return re:IsActivated() and bit.band(r,REASON_COST)~=0
end
```

### DIFF
--- a/field.cpp
+++ b/field.cpp
@@ -1661,6 +1661,11 @@ int32 field::get_release_list(uint8 playerid, card_set* release_list, card_set* 
 				effect* peffect = pcard->is_affected_by_effect(EFFECT_EXTRA_RELEASE_NONSUM);
 				if(!peffect || (peffect->is_flag(EFFECT_FLAG_COUNT_LIMIT) && peffect->count_limit == 0))
 					continue;
+				pduel->lua->add_param(core.reason_effect, PARAM_TYPE_EFFECT);
+				pduel->lua->add_param(REASON_COST, PARAM_TYPE_INT);
+				pduel->lua->add_param(core.reason_player, PARAM_TYPE_INT);
+				if(!peffect->check_value_condition(3))
+					continue;
 				if(ex_list_oneof)
 					ex_list_oneof->insert(pcard);
 				ex_oneof_max = 1;


purerosefallen:
Why not re:IsHasType(0x7f0)

AndreOliveiraMendes:
why not making a diferent function or puting a param for reason in tribute related function?

mercury233:
in case of effect types change, we can just update core instead of scripts

AndreOliveiraMendes:
but how core know it tributing for cost and not other reason?

AndreOliveiraMendes:
i still don't get, don't that still makes you have to update script to check if they use correct function?

mercury233:
it is the answer to "Why not re:IsHasType(0x7f0)" ...

edo9300:
I think this approach is the most unsafe one, the actual core does lots of assumptions in its functions, for example "in this situation, when the function is called there's no way x card could behave in a different way", but then a new card is released that makes this approach invalid and needs lots of old cards to be reworked. One of the most remarkable example are effects that activates when an opponent's monster is attacking, like mirror force, they work on the assumption that if an attack is declared during the opponent's turn, for sure it has been declared by the opponent, thus in the scripts only the turn player is checked. But here it comes the first fail of this reasoning: ancient lamp, a pretty old card, that makes the opponent attack himself, since lots of cards use that check even if they activates when an opponent monster battled with a monster of yours, can activate their effects. We risked even with the card last turn, still not fully implemented, whose effect makes the players enter in a "special battle phase" that behave in the same way as the normal one letting you activate those effects, fortunately konami made the effect in a way so that it's the opponent who declares the attack, but if konami ignored this "basic rule" like they almost always do with those weird cards, it could have made your monster declare the attack, making the opponent unable to activate his effects that activated when the opponent attacks, but you would have been able. So, as you can see we cannot do those assumptions when scripting cards knowing that probably konami will release a card that would make them false. As for this scenario, the less wrong way would be those different functions that, even if in the core just change a parameter, let you make a proper check since when that function is called it'd be sure it's been called for that purpose. A thing I thought about some time ago, was to add an extra parameter to the functions that allows an indefinite number of parameters, like all the group checking/getting ones, this parameter will tell the core how many default parameters have been added before the extra ones, if this approach was used since the beginning, adding extra parameters to those functions wouldn't have required a mass change in all the old scripts, or the needing of adding extra functions, an example is getreleasegroup, which has a "counterpart" getreleasegroupex whose behaviour is exactly the same but with the difference it adds an extra parameter in the core. If that way I described above was used before there wouldn't have been need of all those ad hoc functions that have to be added each time konami releases a card that makes our assumptions false.

PR #147, opened 2018-04-18
Update card::get_link()

Yumekaki:
For "Link Number", official cards reference "リンクＸ (Link X)" while comparing it with a designated value and "リンクマーカーの数 (count of link markers)" while calculating the sum of it or using it as a counter. "Link Number", unlike level or rank, is referencing to the "count of link markers", rather than some sort of card attribute.

Cards that referenced "リンクマーカーの数" instead of "リンクＸ" (18th, April, 2018):
●V−LAN ヒドラ
●ザ・アキュムレーター
●天火の牢獄
●サイバーサル・サイクロン
●バウンドリンク
●ショートヴァレル・ドラゴン

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -963,7 +963,7 @@ uint32 card::get_link() {
 		return 0;
 	uint32 link = 0;
 	uint32 link_marker = get_link_marker();
-	while(link_marker)  
+	while(link_marker > 0)
 	{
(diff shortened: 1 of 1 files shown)


DailyShana:
But there still exist cards that use "リンクＸ (Link X)" directly, so it should be another function to count the number of Link Markers.
If it were necessary, we would add it.

PR #148, opened 2018-04-21
Update Duel.SelectMatchingCard

Aephiex:
当前的ygopro的表现：
1. 无论其他区域是否有足量的符合条件的卡，都会查看卡组。
2. 玩家能清楚地知道卡组中的符合条件的卡各自的位置。
3. 只要没有将卡从卡组取出，就不会洗牌。

应有的表现：
1. 如果其他区域有足量的符合条件的卡，无需查看卡组。
2. 如果查看了卡组，即便没有从卡组将卡取出，也要洗牌。
3. 即便卡组没有符合条件的卡，也可以查看卡组，并洗牌。

----------------------------------------------------------------------------------------------------

How it goes in current ygopro:
1. Will always search the deck regardless of whether there are enough eligible cards in other locations.
2. The player will clearly know the sequence of each eligible card in deck.
3. As long as it hasn't take any cards from the deck, it won't shuffle the deck.

How it should goes:
1. If there are enough eligible cards in other locations, there is no need to search the deck.
2. The deck will be shuffled after the search even if no cards were taken from it.
3. Can still search the deck then shuffle it even if there are no eligible cards in the deck.

### DIFF
--- a/libduel.cpp
+++ b/libduel.cpp
@@ -2226,8 +2226,91 @@ int32 scriptlib::duel_select_matching_cards(lua_State *L) {
 	uint32 location2 = lua_tointeger(L, 5);
 	uint32 min = lua_tointeger(L, 6);
 	uint32 max = lua_tointeger(L, 7);
+	
+	// see if we can search deck
+	int32 can_search_1;
+	int32 can_search_2;
+	if (playerid == self)
+	{
+		int32 can_search_1 = location1 & LOCATION_DECK > 0;
+		int32 can_search_2 = location2 & LOCATION_DECK > 0;
+	}
+	else
+	{
+		int32 can_search_1 = location2 & LOCATION_DECK > 0;
+		int32 can_search_2 = location1 & LOCATION_DECK > 0;
+	}
+	
+	// generate selectable group from deck
 	group* pgroup = pduel->new_group();
+	group* pgroup_sd = pduel->new_group();
+	group* pgroup_od = pduel->new_group();
+	if (can_search_1)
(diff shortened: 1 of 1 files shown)


Aephiex:
https://github.com/Fluorohydride/ygopro/pull/2108

mercury233:
Do you mean that Igknight cards should be able to look at and shuffle deck even if there is no suitable card in deck?

mercury233:
And the player can directly pick card from grave to avoid shuffling the deck?

Aephiex:
@mercury233 目前的ygopro中，即使玩家最终决定从墓地拿取点火骑士，玩家将仍然有机会记住每个符合条件的点火骑士在自己卡组中的位置。如果玩家希望自己抽到的下一张牌是点火骑士，玩家发现自己卡组顶就是想要的点火骑士的话，玩家完全可以从墓地拿取点火骑士。然而在OCG中，如果你知道卡顶是点火骑士，说明你看了卡组，那么无论最后你从哪里拿取点火骑士，卡组都要洗牌。

purerosefallen:
To implement what you wanted needs many steps, including add a new processor. The code in your PR may not really implement it.

edo9300:
Those hardcodings should be avoided as much as possible, this shuffle should be handled in the script, not in the core, otherwise it could cause troubles with other cards, even custom, that clearly say, for example, that the deck isn't shuffled, also if you do this, then you should update every other function that let's you select from the deck

Aephiex:
@edo9300 
Fine. I will realize this feature by mass altering card script.

mercury233:
so another utility function?

Aephiex:
https://github.com/Fluorohydride/ygopro-scripts/pull/1022
https://github.com/Fluorohydride/ygopro-scripts/pull/1023

PR #151, opened 2018-04-23
Effect.UseCountLimit & Effect.CheckCountLimit

Aephiex:
Adds two interface functions.


purerosefallen:
The Effect.UseCountLimit should be return 0 in the end since it has no returns.

Aephiex:
https://github.com/Fluorohydride/ygopro-scripts/pull/1025

PR #156, opened 2018-05-04
change announce card behavior

mercury233:
Some user may use custom expansion databases, and declare card that only exists in those databases.
Now because the client don't actually support MSG_RETRY, those declaration will break the duel.
So we should make it able to actually retry when the user declare illegal card.

> !system 1421 宣言的卡不符合条件，或无法被主机识别。

### DIFF
--- a/playerop.cpp
+++ b/playerop.cpp
@@ -1010,22 +1010,26 @@ int32 field::announce_card(int16 step, uint8 playerid, uint32 ttype) {
 		return FALSE;
 	} else {
 		int32 code = returns.ivalue[0];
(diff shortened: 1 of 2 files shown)


mercury233:
@purerosefallen it uses same `field::announce_card`

purerosefallen:
Does this change affect client? Can it be compatible without updating client, except for replay?

mercury233:
no, it just send MSG_ANNOUNCE_CARD again

purerosefallen:
Remember to add this line into strings.conf

PR #160, opened 2018-05-10
update EFFECT_DISABLE_FIELD for Iron Dragon Tiamaton

mercury233:
Now if the sequence of _Iron Dragon Tiamaton_ is moved, the disabled zones aren't changed.
I think this may fix it.

```lua
	local e4=Effect.CreateEffect(c)
	e4:SetType(EFFECT_TYPE_FIELD)
	e4:SetRange(LOCATION_MZONE)
	e4:SetCode(EFFECT_DISABLE_FIELD)
	e4:SetProperty(EFFECT_FLAG_REPEAT)
	e4:SetOperation(c46247282.disop)
	c:RegisterEffect(e4)
```

### DIFF
--- a/processor.cpp
+++ b/processor.cpp
@@ -4988,7 +4988,7 @@ int32 field::refresh_location_info(uint16 step) {
 		filter_field_effect(EFFECT_DISABLE_FIELD, &eset);
 		for (int32 i = 0; i < eset.size(); ++i) {
 			value = eset[i]->get_value();
-			if(value) {
+			if(value && !eset[i]->is_flag(EFFECT_FLAG_REPEAT)) {
 				player[0].disabled_location |= value & 0x1f7f;
 				player[1].disabled_location |= (value >> 16) & 0x1f7f;
 			} else


purerosefallen:
I think there could be a better solution.
To check EFFECT_FLAG_REPEAT after running the operation. If this flag exists, the value of the effect will not be overwritten.
Then, the disabled field could be from the effect value (A function) in the script.

mercury233:
I don't know why but all `EFFECT_DISABLE_FIELD` don't use value from script.

purerosefallen:
Because it requires players to choose a zone to disable in each operation. In addition, values are occupied for storing the selected disabled zone.

purerosefallen:
will https://github.com/Fluorohydride/ygopro-scripts/commit/773e9017c6b3ab1f2ec213da42ef5274f8c039c9 work?

destdev:
I've just tested this script change and it seems to work. I was freely able to move that monster around with Senet Switch every turn and all disabled zones moved with him. The monster was moved multiple times, all the disabled zoned moved with it, properly along all columns. Unless I'm missing any method (normal/special summon, cards moving around like with Senet Switch or something, cards that just "place" card on the field from outside like the True Draco spell/trap searchers) I don't think there is any way to move to these fields. So, I guess the suggested script way should be fine?
unbenannt

mercury233:
image


set value to 0 will cause the disabled field actually become 0 when equipping monster from hand to field.

PR #163, opened 2018-05-19
update Card.IsLevel/Rank/Link

nekrozar:
Update this: We can use `Card.IsLevel/Rank/Link(X,Y,Z,...)`.


purerosefallen:
Are you sure every time you run `get_level()`, it returns the same?
I suggest to store it at first, and then compare the value with the given levels.

nekrozar:
https://github.com/Fluorohydride/ygopro-core/pull/163/commits/369454667cf20068b12a98f1873d5d79b7f1f178 updated.

PR #164, opened 2018-05-22
add Card.IsAttack/Defense

nekrozar:
Update this: We can use `Card.IsAttack/Defense(X,Y,Z,...)`.


purerosefallen:
Did you consider the attack `?` ?
When you say `c:IsAttack(-1)` or any other negative values, it should return true for any `?` attack monsters.
Also the link monsters for `Card.IsDefense` should also be considered.

nekrozar:
I checked, `?` attack monsters equal `c:IsAttack(-2)`.

nekrozar:
I checked, _Return of the Monarchs_ operate correctly.

PR #167, opened 2018-06-14
update move_to_field

mercury233:
skip `PROCESSOR_SELECT_PLACE` if `zone` only contain 1 avail place.


purerosefallen:
I think there should be a better way to detect a single bit, like `v & (v-1) == 0`.

mercury233:
but we need to get the seq too

PR #169, opened 2018-06-24
update Card.IsReleasableByEffect

nekrozar:
Update This: If _Lair of Darkness_ has on the field, _Cocoon of Ultra Evolution_ cannot release opponent monster.


DailyShana:
I think `EFFECT_EXTRA_RELEASE_NONSUM` should be removed here because it's no possible that the monster assigned by effect cannot be released by that effect.

PR #171, opened 2018-07-12
small name change

AndreOliveiraMendes:
updated the name.
related with https://github.com/Fluorohydride/ygopro-scripts/pull/1075


mercury233:
Why is the new name better?

edo9300:
The first name is not correct grammatically, also could be mistaken with the summon from the extra deck, this one is grammatically correct and more intuitive

purerosefallen:
Why not making it return the count

AndreOliveiraMendes:
or probaly either the bool and count?

DailyShana:
Literally, if CheckAdditionalSummon returns true, it should mean the player have not use the chance of extra normal summon and can activate those effects.

mercury233:
https://github.com/Fluorohydride/ygopro-core/commit/ce33757015da7334e6883589bc6288a64823f1c0 merged

purerosefallen:
I'm going to remove this function and add a param to Duel.IsPlayerCanSummon.

PR #172, opened 2018-07-19
fix for EFFECT_CANNOT_DISEFFECT

AndreOliveiraMendes:
related with the issue https://github.com/Fluorohydride/ygopro/issues/2122


DailyShana:
Sorry I change the ideal since "effect cannot be disabled" is not effect that affect cards.
It is fixed in https://github.com/Fluorohydride/ygopro-scripts/commit/bb86089c53615d45fca86dc109d9b80de42db51f#diff-ed7a908be9c6eddb7f540636114e0aff

PR #176, opened 2018-08-10
small mend for undefined behaviour

DyXel:
Under C++ standard, type punning is
Undefined Behaviour and strict aliasing rules
forbids writing data to an array of different type,
this is a small change that attempts to fix this issue
by just using std::memcpy, which is effectively
optimized by modern compilers.

### DIFF
--- a/duel.cpp
+++ b/duel.cpp
@@ -116,17 +116,17 @@ void duel::restore_assumes() {
 	assumes.clear();
 }
 void duel::write_buffer32(uint32 value) {
-	*((uint32*)bufferp) = value;
+	std::memcpy(bufferp, &value, sizeof(value));
 	bufferp += 4;
 	bufferlen += 4;
 }
 void duel::write_buffer16(uint16 value) {
-	*((uint16*)bufferp) = value;
+	std::memcpy(bufferp, &value, sizeof(value));
 	bufferp += 2;
 	bufferlen += 2;
 }
 void duel::write_buffer8(uint8 value) {
-	*((uint8*)bufferp) = value;
+	std::memcpy(bufferp, &value, sizeof(value));
 	bufferp += 1;
 	bufferlen += 1;
 }


DailyShana:
https://github.com/Fluorohydride/ygopro-core/blob/master/duel.cpp#L15
The header included here is <memory.h>, so `std::` is unnecessary

DailyShana:
In fact, `-fno-strict-aliasing` is already set here https://github.com/Fluorohydride/ygopro/blob/366ff6e44160c8766c65edc2e85b3ee336f95f32/premake4.lua#L30
and `char*`(`bufferp` here) can alias to other types as an exception of strict aliasing rule.

DyXel:
`std::` is the namespace from where all functions that comes from within the C++ standard library reside, and should be used accordingly to identify those functions. In this very specific PR im using `std::memcpy` and not `memcpy`, which is the C++ standard way of doing so.

> In fact, -fno-strict-aliasing is already set here

I will not explain in detail how bad this flag is; It affects massively the performance of the entire program (by removing certain optimizations that would break [the now existing] aliasing rules), it should only be used by people that have specific case scenarios where its strictly needed, and even if all of that is omitted some compilers do not even support it, for more than enough reasons. We could just solve this altogether by doing the right thing and there would be no need for this horrendous flag.

>and char*(bufferp here) can alias to other types as an exception of strict aliasing rule.

While this part might be true, its only a half-truth. To properly bypass the strict aliasing exception, you will need to let the compiler know about your intentions, which in this case would be by doing `reinterpret_cast` from a type to a type that satisfies `AliasedType`, which the original code doesn't do either. Even cppreference.com suggest programmers on using std::memcpy for such situations:

`When it is needed to interpret the bytes of an object as a value of a different type, std::memcpy or std::bit_cast (since C++20)can be used`
`Where strict aliasing prohibits examining the same memory as values of two different types, std::memcpy may be used to convert the values.`

from: https://en.cppreference.com/w/cpp/language/reinterpret_cast and https://en.cppreference.com/w/cpp/string/byte/memcpy respectively

Fixing and refactoring code in this state should be the priority number one before doing further additions IMO, we can begin to do this by solving issues like this one, one by one.

salix5:
@DyXel 
We have removed all type punning in this repo now.
Thank you!

PR #177, opened 2018-08-12
Fixed GetLabelObject to work with the last changes made by shana

edo9300:


libeffect.cpp
```diff
@@ -343,11 +343,11 @@ int32 scriptlib::effect_get_label_object(lua_State *L) {
 		return 1;
 	}
 	int32 type = *(int32*)peffect->label_object;
-	if(type == 1)
+	if(check_data_type(L, -1, "Card"))
```
DailyShana: the top of the stack is the effect parameter, not the label_object

PR #181, opened 2018-08-20
update attacking process

DailyShana:
Use another `attack_announce_count` for announce count in scripts.
`announce_count` is only used in ocgcore. 

When selecting attack target, `announce_count` don't increase. It increases only when the attacking is passed or stopped.
So that the behavior of `get_attack_target` is the same when getting attackable monsters and selecting attack target.
And merge `confirm_attack_target` into `get_attack_target` via a parameter `select_target`.

All these can reduce much repeated code in `get_attack_target` and `confirm_attack_target` and make it easier to maintain them.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -60,6 +60,14 @@ void card::attacker_map::addcard(card* pcard) {
 	auto pr = emplace(fid, std::make_pair(pcard, 0));
 	pr.first->second.second++;
 }
+uint32 card::attacker_map::findcard(card* pcard) {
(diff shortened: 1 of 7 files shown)


DailyShana:
Make the logic in `get_attack_target` more clear.
And for https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=19387&keyword=&tag=-1

DailyShana:
move to #182

PR #185, opened 2018-09-03
Updated api to allow script loading from buffer

edo9300:


DailyShana:
extra message is nothing to do with duel process, it should not be a part of ocgapi.

edo9300:
> extra message is nothing to do with duel process, it should not be a part of ocgapi.

this is not only related to the extra message, with this change, you could avoid making the api call the interpreter when loading a script and directly call some scripts without needing to open files and other things

PR #186, opened 2018-09-22
fix undefined behavior caused by lua_yield

DailyShana:
There is a `longjmp` behind `lua_yield` (jump to `setjmp` behind `lua_resume`), so if there are local variables that have non-trivial destructor used before `lua_yield`, the behavior is undefined.

Another solution is to build lua library as C++ code. `lua_resume`/`lua_yield` will use `try{}catch(...)`/`throw`, so destructor will be called normally.

I think this is the reason of these issues: Fluorohydride/ygopro#1854 Fluorohydride/ygopro#1682 #141

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -1534,11 +1534,6 @@ int32 card::get_old_union_count() {
 	}
 	return count;
 }
-void card::xyz_overlay(card* pcard) {
-	card_set materials;
-	materials.insert(pcard);
-	xyz_overlay(&materials);
-}
 void card::xyz_overlay(card_set* materials) {
 	if(materials->size() == 0)
 		return;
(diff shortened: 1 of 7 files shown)


InvisibleMoon:
That's cool! Could you offer some detail explanation of what actually happened, in my mind, it seems that some variables were destructed in the C++ context or the stack was broken.

DailyShana:
for example, when a monster is going to special summon itself,
the call stack is roughtly like this: (read from bottom to top)
```
LUAI_THROW (longjmp)  // goto LUAI_TRY, no guaranteed stack unwinding
...
lua_yield // field::card_set cset used before this
scriptlib::duel_special_summon (lua function Duel.SpecialSummon)
...
LUAI_TRY (setjmp)
...
lua_resume
interpreter::call_coroutine
field::execute_operation
...
```
so the destructor of `cset` maybe not called, causing sth wrong.

PR #187, opened 2018-09-25
update EFFECT_FLAG_REPEAT of get atk/def

DailyShana:
fix the atk effect of _The Wicked Avatar_.
Let _Clear Vice Dragon_ equip _Moon Mirror Shield_, attack _The Wicked Avatar_. After applying the effect of _Moon Mirror Shield_, the atk of _Clear Vice Dragon_ is 1000 but the atk of _The Wicked Avatar_ is only 300, which should be 1100.

The only one issue of this solution is the possibility of infinite loop. We should detect it and return `temp.attack` instead of calculate `get_value` again.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -667,9 +667,9 @@ int32 card::get_attack() {
 			temp.attack = get_defense();
 		}
 		for(int32 i = 0; i < effects_atk_r.size(); ++i) {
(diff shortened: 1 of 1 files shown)


purerosefallen:
Could you please make PRs from your own fork?

PR #191, opened 2018-10-17
update summon proc

DailyShana:
when monster use `EFFECT_SUMMON_PROC`, for extra opportunity of normal summon count, it ignores the zone and releasable parameter of extra summon count, so the scripts of that kind of monsters need update.
This pull request set the parameter internally to avoid updating scripts.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -2868,7 +2868,6 @@ int32 card::check_cost_condition(int32 ecode, int32 playerid, int32 sumtype) {
 	}
 	return TRUE;
 }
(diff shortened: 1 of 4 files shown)


edo9300:
What are the cards that will be fixed with this PR? Also in what scenarios?

DailyShana:
e.g. using extra opportunity of normal summon count given by _Knightmare Goblin_ to summon _Ehther the Heavenly Monarch_, the _Knightmare Goblin_ cannot be selected as tribute

PR #192, opened 2018-10-18
multiple effect label

DailyShana:
It will be useful for some cards like _Tuner's High_.

### DIFF
--- a/effect.cpp
+++ b/effect.cpp
@@ -41,7 +41,7 @@ effect::effect(duel* pd) {
 	reset_flag = 0;
 	count_code = 0;
 	category = 0;
-	label = 0;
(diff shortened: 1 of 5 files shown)


purerosefallen:
`e:SetLabel(c:GetAttack() | c:GetDefense()<<16 | c:GetLevel()<<24 | ...)` may be enough
Shell we put the label to int64?

purerosefallen:
@edo9300 why not a table? table can do anything

edo9300:
@purerosefallen if a table can do anything, a function can do even more

purerosefallen:
@edo9300 `{f=function() ..... end}` ~~So a table is better~~

PR #201, opened 2018-12-27
Coding convention fix (like unify) for another project

ZisIsNotZis:
I am working on [a project](https://github.com/ZisIsNotZis/ygoDocGen) that generates Lua Documentation by automatically inspecting CPP files in `ygopro-core` . The current program kind of strongly relies on spacing. While debugging I found some part of the `ygopro-core` source code forgot to add space when type casting. I figure it is better to directly fixing the upstream rather than creating another branch or doing complicated checking in the script. Therefore here is the fix.

It basically changes all 

    *(xxx**)lua_touserdata

to

    *(xxx**) lua_touserdata

(with a space in the middle). Most part of the code already have space. Some don't.

And another fix is changing from a `L,3` to `L, 3`


DailyShana:
unlike python, whitespace is not important in c++, so please improve your code to avoid dependency on spaces.

PR #209, opened 2019-03-10
changed assume to a map

AndreOliveiraMendes:
changed assume to a map, so it allow more than 2 diferent properties to be assumed at any time, may help with the [issue](https://github.com/Fluorohydride/ygopro-scripts/issues/741) opened there.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -94,8 +94,6 @@ card::card(duel* pd) {
 	spsummon_counter_rst[0] = spsummon_counter_rst[1] = 0;
 	unique_code = 0;
 	unique_fieldid = 0;
-	assume_type = 0;
-	assume_value = 0;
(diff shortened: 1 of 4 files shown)


DailyShana:
That issue is due to assumes can't be applied in the hand written synchro process in scripts, nothing to do whether assumes are maps or not.

purerosefallen:
What is `lua_tonumberint`?
If it is a function in @edo9300's fork, please change it back.

AndreOliveiraMendes:
whatever, the issue of assume only taking last propery still exist, like, you can't assume race and type at same time and this may cause conflict one day.
also @purerosefallen didn't saw it, alredy changed.

purerosefallen:
repo gone, and the merge can't be done, and it's too outdated

PR #210, opened 2019-03-13
add more triggering states

mercury233:


purerosefallen:
How about links and link markers?

mercury233:
_let us wait some effects can change it_

PR #216, opened 2019-05-26
Merge MSG_ANNOUNCE_CARD with MSG_ANNOUNCE_CARD_FILTER

edo9300:
Now they won't be treated as separated messages anymore, the old announce card will be like an announce card filter with no opcodes, while in case it had the type parameter, the is type opcode will be sent


DailyShana:
I think then `Duel.AnnounceCard` and `Duel.AnnounceCardFilter` should also be merged together.

purerosefallen:
How about script updates?

edo9300:
@purerosefallen you're right, updating the 2 only cards that use the filter function would be way too difficult

edo9300:
Ok, @DailyShana , i merged the 2 functions into one, if only 1 aurgment is passed, it'll use the "default" types for declaration, if exactly 2 are passed, then that means it's the "custom" type for declaration, otherwise it'll read the params as the various opcodes

DailyShana:
>if exactly 2 are passed, then that means it's the "custom" type for declaration, 

why not change this case to opcodes in scripts

edo9300:
> > if exactly 2 are passed, then that means it's the "custom" type for declaration,
> 
> why not change this case to opcodes in scripts

what do you mean?

DailyShana:
change `Duel.AnnounceCard(tp,type)` to `Duel.AnnounceCard(tp,type,OPCODE_ISTYPE)` in scripts
and the special treatment for 2 parameters can be removed

edo9300:
Done

purerosefallen:
I think compating the 2 param is needed. Because there are other pre-release or custom/manga cards in other communities. Also `Duel.AnnounceCardFilter` can be left for compating.
I don't want to see players complaining why their cards not working.

DailyShana:
@purerosefallen 
regex replace:
find: `AnnounceCard\((.*),(.*)\)`
replace: `AnnounceCard($1,$2,OPCODE_ISTYPE)`

PR #222, opened 2019-06-21
use lua continuation to give return value of operation functions

DailyShana:
simplify the returning value of these lua functions by removing redundant PROCESSOR_***_S

reference: https://www.lua.org/manual/5.3/manual.html#4.7


edo9300:
Hmm interesting thing, this will make the code cleaner too

PR #224, opened 2019-06-26
add EFFECT_TYPE_TARGET

DailyShana:
for effect affecting continuous targets, parallel to `EFFECT_TYPE_EQUIP`

Note:
`EFFECT_TYPE_TARGET` is only for effects that **affect** continuous targets. The continuous spell/trap taking continuous targets on field when activating is of this kind, which is different from monster effects that taking continuous targets, see Fluorohydride/ygopro-scripts#1203

Sample:
```lua
local e3=Effect.CreateEffect(c)
e3:SetType(EFFECT_TYPE_TARGET)
e3:SetCode(effect_code)
e3:SetRange(location) --where to apply
--[[other settings:
e3:SetCondition
e3:SetTarget --additional condition of affected continuous target
e3:SetProperty
e3:SetValue
...
--]]
c:RegisterEffect(e3)
```

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -2244,10 +2244,19 @@ void card::cancel_card_target(card* pcard) {
 	}
 }
 void card::clear_card_target() {
-	for(auto& pcard : effect_target_owner)
+	for(auto& pcard : effect_target_owner) {
 		pcard->effect_target_cards.erase(this);
+		for(auto& it : pcard->target_effect) {
+			if(it.second->is_disable_related())
+				pduel->game_field->add_to_disable_check_list(this);
+		}
+	}
 	for(auto& pcard : effect_target_cards) {
 		pcard->effect_target_owner.erase(this);
+		for(auto& it : target_effect) {
+			if(it.second->is_disable_related())
+				pduel->game_field->add_to_disable_check_list(pcard);
+		}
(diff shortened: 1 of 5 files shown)


effect.cpp
```diff
@@ -96,7 +96,7 @@ int32 effect::is_available() {
 		if(powner == phandler && !is_flag(EFFECT_FLAG_CANNOT_DISABLE) && phandler->get_status(STATUS_DISABLED))
 			return FALSE;
 	}
-	if (type & EFFECT_TYPE_EQUIP) {
+	if(type & EFFECT_TYPE_EQUIP) {
```
DailyShana:
```suggestion
	if (type & EFFECT_TYPE_EQUIP) {
```

card.cpp
```diff
@@ -1780,33 +1791,40 @@ void card::remove_effect(effect* peffect) {
...
+		target_effect.erase(it);
+		if(!effect_target_cards.empty())
+			check_target = effect_target_cards;
+		else
+			check_target.clear();
 	} else if (peffect->type & EFFECT_TYPE_XMATERIAL) {
 		xmaterial_effect.erase(it);
-		if (overlay_target)
-			check_target = overlay_target;
+		if(overlay_target)
```
DailyShana:
```suggestion
		if (overlay_target)
```

PR #227, opened 2019-07-07
rebuild desc hint during shuffle hand

DailyShana:
fix Fluorohydride/ygopro#1876

### DIFF
--- a/field.cpp
+++ b/field.cpp
@@ -910,17 +910,6 @@ void field::shuffle(uint8 playerid, uint8 location) {
 			core.shuffle_hand_check[playerid] = FALSE;
 			return;
 		}
-		for(auto& pcard : svector) {
-			for(auto& i : pcard->indexer) {
-				effect* peffect = i.first;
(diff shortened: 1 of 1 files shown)


field.cpp
```diff
@@ -910,6 +910,17 @@ void field::shuffle(uint8 playerid, uint8 location) {
...
+			for(auto& i : pcard->indexer) {
+				effect* peffect = i.first;
+				if(peffect->is_flag(EFFECT_FLAG_CLIENT_HINT) && !peffect->is_flag(EFFECT_FLAG_PLAYER_TARGET)) {
+					pduel->write_buffer8(MSG_CARD_HINT);
+					pduel->write_buffer32(pcard->get_info_location());
+					pduel->write_buffer8(CHINT_DESC_REMOVE);
+					pduel->write_buffer32(peffect->description);
+				}
+			}
+		}
```
DailyShana: this part can be omitted if let client clear all desc hint when receiving MSG_SHUFFLE_HAND

PR #231, opened 2019-07-14
update Duel.SelectDisableField for extra mzone

purerosefallen:


DailyShana:
it's better to allow selecting extra mzone automatically when `filter` parameter contains bits for extra mzone

purerosefallen:
@DailyShana `filter` is a bit for the zone NOT selectable. So it's not very proper to use `filter` for extra zones.

DailyShana:
it natural to use `filter` to decide which zone is selectable, for cards that cannot select extra moster zone, just make a little modify

purerosefallen:
*Making announcements is painful*

PR #232, opened 2019-07-14
update Duel.SelectDisableField

DailyShana:
allow selecting extra monster zones, only when selecting monster zones of both players, as empty extra monster zones do not belong to neither player according to the rule


nekrozar:
should we update _Excode Talker_ and so on?
I confirmed, _Excode Talker_ can select extra monster zone.

DailyShana:
@nekrozar Yes, I'll update them after merging this.

PR #233, opened 2019-07-14
add chint to opponent's hand

nekrozar:
automaticity https://github.com/Fluorohydride/ygopro-scripts/commit/ea0e0cb76cb351ea68c37c84f368fdf238b730d2

### DIFF
--- a/operations.cpp
+++ b/operations.cpp
@@ -408,6 +408,16 @@ int32 field::draw(uint16 step, effect* reason_effect, uint32 reason, uint8 reaso
 				shuffle(playerid, LOCATION_HAND);
 			}
 			for (auto& pcard : *drawed_set) {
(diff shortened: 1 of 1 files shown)


DailyShana:
maybe it can be done in `add_card()`?

nekrozar:
@DailyShana If these code moved in `add_card`, ygopro crashed.

edo9300:
@nekrozar what kind of crash? Cannot you follow the call stack and find out the issue?

edo9300:
The crash is happening because MSG_CARD_HINT is being written by add_effect while MSG_MOVE is still being written

nekrozar:
This problem happen in `ClientField::AddCard`.

@edo9300 thanks for find and confirm.

PR #236, opened 2019-08-22
add Card.GetSummonSequence

nekrozar:
for "monster is Special Summoned to a zone" effect.

### DIFF
--- a/card.h
+++ b/card.h
@@ -133,6 +133,7 @@ public:
 	query_cache q_cache;
 	uint8 owner;
 	uint8 summon_player;
+	uint8 summon_sequence;
 	uint32 summon_info;
 	uint32 status;
 	sendto_param_t sendto_param;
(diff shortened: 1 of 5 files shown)


DailyShana:
it can be solved in #223
and it is meaningless to think summon infomation of monsters not on field, it causes some problem like Fluorohydride/ygopro#1737

nekrozar:
> and it is meaningless to think summon infomation of monsters not on field, it causes some problem like Fluorohydride/ygopro#1737

Sorry, I couldn't understand this.

DailyShana:
"special summoned from extra deck" is meaningless for monsters in grave.
first special summon some monster from extra, then send it to grave. In other turn use Monster Reborn to special summon it from grave, then The Monarchs Stormforth cannot be activated this turn in ygopro.
this is because that monster has summon_location==LOCATION_EXTRA even in grave

nekrozar:
thanks.
these pull requests close.

PR #241, opened 2019-10-24
add card status detection when checking summonable linked zone

purerosefallen:
May solve https://github.com/Fluorohydride/ygopro/issues/2233


DailyShana:
Instead of add an extra parameter to `is_link_marker`, I think add a new `is_link_marker_on_field` is better.

DailyShana:
After consideration, I think `is_link_marker` should always check status summing, etc when the link monster is on field.

DailyShana:
fix in #251

PR #244, opened 2019-11-30
add Card.IsCanOverlay

nekrozar:
> https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=22761
> 「永の王 オルムガンド」のモンスター効果によってエクシーズ素材とするカードを選ぶ際に、モンスタートークンを選ぶ事はできません。

but, _Jormungandr, Generaider Boss of Eternity_ can select equipping tokens.

> https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=7967
> 「禁止令」の効果が適用されている場合、そのカード名が宣言されているモンスターを「十二獣の方合」の効果によってデッキから選び、エクシーズ素材とする事自体ができません。

and, forbidden check by default.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -3700,6 +3700,15 @@ int32 card::is_capable_be_effect_target(effect* peffect, uint8 playerid) {
 	}
 	return TRUE;
 }
+int32 card::is_capable_overlay(uint8 playerid) {
+	if(data.type & TYPE_TOKEN)
+		return FALSE;
+	if(!(current.location & LOCATION_ONFIELD) && is_status(STATUS_FORBIDDEN))
+		return FALSE;
+	if(current.controler != playerid && !is_capable_change_control())
+		return FALSE;
+	return TRUE;
+}
 int32 card::is_can_be_fusion_material(card* fcard, uint32 summon_type) {
 	if(is_status(STATUS_FORBIDDEN))
 		return FALSE;
--- a/card.h
+++ b/card.h
@@ -342,6 +342,7 @@ public:
 	int32 is_control_can_be_changed(int32 ignore_mzone, uint32 zone);
 	int32 is_capable_be_battle_target(card* pcard);
 	int32 is_capable_be_effect_target(effect* peffect, uint8 playerid);
+	int32 is_capable_overlay(uint8 playerid);
 	int32 is_can_be_fusion_material(card* fcard, uint32 summon_type);
 	int32 is_can_be_synchro_material(card* scard, card* tuner = 0);
 	int32 is_can_be_ritual_material(card* scard);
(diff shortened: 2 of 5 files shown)


DailyShana:
Card.IsCanBeOverlayed is better

mercury233:
Overlay something to something, so "is can be overlaid" is worse for material card

DailyShana:
https://dictionary.cambridge.org/dictionary/english/overlay
The tiny carvings were overlaid with silver

so it should be "the material card is overlaid with the xyz monster"

mercury233:
how about "IsCanBeAttached"

DailyShana:
attach is also a verb that can have double object
attach A to B
A is attached to B
B is attached A
so it should be `IsCanBeAttachedTo`

But I suggest `c:IsCanBeXyzMaterial(xc,REASON_EFFECT)` for this kind of effect,
`c:IsCanBeXyzMaterial(xc,REASON_XYZ[+REASON_MATERIAL])` for xyz summon.

mercury233:
using `IsCanBeXyzMaterial` at two different case isn't a good idea

edo9300:
This approach is better, because it has a clearer usage of the function, adding future proofing for any kind of effects or ruling changes only for a specific scenario

mercury233:
It is annoying to use a parameter for distinguishing 2 totally different function.

DailyShana:
not different, they are all about whether a card can be xyz material, one for becoming material by xyz summon, the new one is for becoming material by effect.

mercury233:
But the checks are different

purerosefallen:
@DailyShana Why are you always thinking of changing the function name?
Last time `EFFECT_MUST_USE_MZONE`'s name were also commented by you suggesting for a name change. And it happened again this time.
It should be a common sense that the function names in this project should be decided by the creator of the functions, rather than "fitting" the English grammar.
In addition, frequently changing function name just for grammar would result in many workings in the downstream projects, like `ygopro-pre-script`. Please stop thinking about making those changes. Thanks.

DailyShana:
> @DailyShana Why are you always thinking of changing the function name?

we should alway use proper, intuitive names instead of password-like names
> Last time `EFFECT_MUST_USE_MZONE`'s name were also commented by you suggesting for a name change. And it happened again this time.

this time is completely because of this pull request is merged before discussion
> It should be a common sense that the function names in this project should be decided by the creator of the functions, rather than "fitting" the English grammar.

never heard such common sense
> In addition, frequently changing function name just for grammar would result in many workings in the downstream projects, like `ygopro-pre-script`. Please stop thinking about making those changes. Thanks.

not frequently, each name is changed at most once, and discussion before decision do not affect downstream and this time is just because this pull request is merged before review.

mercury233:
_generally nobody know if you reviewed any pull request or not, because you don't give any reply_

DailyShana:
> But the checks are different

maybe this point make sense, there are also two functions about being releasable by summon and by effect.

PR #248, opened 2019-12-10
update Duel.SSet

mercury233:
always confirm set cards at default, even if there is only 1 card


nekrozar:
Should we update scripts?
https://github.com/Fluorohydride/ygopro-core/blob/master/operations.cpp#L2481
_Dust Tornado_'s setted card was confirmed.
And many cards call `MSG_CONFIRM_CARDS` twice.

mercury233:
yes, I'll do this later

PR #257, opened 2019-12-19
add Duel.GetSZoneCount

nekrozar:
If no spell&trap zone, _Brotherhood of the Fire Fist - Rooster_ can send field spell to grave.

### DIFF
--- a/field.cpp
+++ b/field.cpp
@@ -773,6 +773,46 @@ uint32 field::get_linked_zone(int32 playerid) {
 	}
 	return zones;
 }
+int32 field::get_zone_count(card* pcard, uint8 playerid, uint8 location, uint8 uplayer, uint32 reason, card* mcard, group* mgroup, uint32 zone, uint8 ex, uint32* list) {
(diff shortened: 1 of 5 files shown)


DailyShana:
I hope you can only focus on szone count and do not change others

nekrozar:
It is better to unify those functions.
If those functions unified, some cards (e.g. _The Weather Painter Sun_) can use new check function (e.g. `Card.IsAbleToField`).

DailyShana:
I don't think there is any improvement for _The Weather Painter Sun_

nekrozar:
some to field effects cannot check perfectly.
so, it is better to add `Card.IsAbleToField`.

DailyShana:
if some card is going to mzone, check the count of mzone
if some card is going to szone, check the count of szone
even if some card is going to mzone or szone, the condition is just mzone>0 or szone>0

nekrozar:
`field::get_zone_count` check mzone or szone only.

PR #258, opened 2019-12-21
fix owner_player of continuous effect of Xyz material

mercury233:
Fix this: The count limit of `EVENT_BE_MATERIAL` effect won't work when the reason is Xyz, and the `tp` param will be `PLAYER_NONE` in script.

Example:
https://github.com/Fluorohydride/ygopro-scripts/blob/master/c63821877.lua#L30


DailyShana:
need further review, EVENT_BE_MATERIAL and the controller of xyz material

DailyShana:
We should use EFFECT_FLAG_EVENT_PLAYER for this kind of effect

PR #261, opened 2019-12-25
update new rule

mercury233:


field.cpp
```diff
@@ -709,6 +708,11 @@ int32 field::get_spsummonable_count_fromex_rule4(card* pcard, uint8 playerid, ui
 		if(!is_location_useable(playerid, LOCATION_MZONE, 6))
 			flag |= 1u << 6;
 	}
+	uint32 linked_zone = 0;
```
DailyShana: even if duel_rule is 4, it include non-linked extra monster zones
```suggestion
	uint32 allowed_zone = 0;
```

field.cpp
```diff
@@ -709,6 +708,11 @@ int32 field::get_spsummonable_count_fromex_rule4(card* pcard, uint8 playerid, ui
...
+	if(core.duel_rule >= 5 && pcard && (pcard->get_type() & TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ) && pcard->is_position(POS_FACEDOWN))
```
DailyShana:
```suggestion
	if(core.duel_rule >= 5 && pcard && (pcard->data.type & TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ) && !((pcard->data.type & TYPE_PENDULUM) && pcard->is_position(POS_FACEUP)))
```
compatible to
https://github.com/Fluorohydride/ygopro-core/blob/c87a077663dd3944e15269f9015708af00c343ef/card.cpp#L3099

field.cpp
```diff
@@ -773,6 +777,15 @@ uint32 field::get_linked_zone(int32 playerid) {
 	}
 	return zones;
 }
+uint32 field::get_fsx_zone_fromex_rule5(int32 playerid) {
```
DailyShana: a rule5 include rule 1~5?

field.cpp
mercury233: what about `rule_zone`

field.cpp
mercury233: > エクストラデッキの**裏側表示**の融合・シンクロ・エクシーズモンスターは
自分のメインモンスターゾーンにも特殊召喚できる
https://www.yugioh-card.com/japan/howto/masterrule2020/

edo9300:
I think this approach will require a script update like the one I had to do on Percy with the usage of GetLocationCountFromEx https://github.com/Ygoproco/Live2017Links/pull/2323, as the card parameter is needed when checking with that function for the correct zones to be returned.

field.cpp
```diff
@@ -773,6 +773,16 @@ uint32 field::get_linked_zone(int32 playerid) {
 	}
 	return zones;
 }
+uint32 field::get_rule_zone_fromex(int32 playerid, card* pcard) {
+	if(core.duel_rule >= 4) {
+		if(core.duel_rule >= 5 && pcard && pcard->is_position(POS_FACEDOWN) && (pcard->data.type & TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ))
```
edo9300: Better go with the check for the negation of the opposite, so not link and not faceup pendulum, to avoid having to update it in case of something new, it even make another function for the card object to keep things clear

PR #262, opened 2019-12-25
Make 2nd parameter in GetLocationCountFromEx optional

edo9300:
It wasn't even really used in scripts, and the only times it was, tp was set as parameter, even of that isn't really the same as reason player on some cases, thus adding also the possibility of wrong interactions in the long run. Since there's no way to retrieve the reason effect (and even if there was it would have been kinda wasting to call a function to pass it as default parameter) it's better to make it optional

### DIFF
--- a/libduel.cpp
+++ b/libduel.cpp
@@ -1869,7 +1869,7 @@ int32 scriptlib::duel_get_location_count_fromex(lua_State *L) {
 		return 0;
 	duel* pduel = interpreter::get_duel_info(L);
 	uint32 uplayer = pduel->game_field->core.reason_player;
-	if(lua_gettop(L) >= 2)
+	if(lua_gettop(L) >= 2 && !lua_isnil(L, 2))
 		uplayer = lua_tointeger(L, 2);
 	bool swapped = false;
 	card* mcard = 0;


mercury233:
isn't tp shorter than nil?------------------ 原始邮件 ------------------
发件人: "Edoardo Lolletti"<notifications@github.com>
发送时间: 2019年12月25日(星期三) 晚上6:25
收件人: "Fluorohydride/ygopro-core"<ygopro-core@noreply.github.com>;
抄送: "Subscribed"<subscribed@noreply.github.com>;
主题: [Fluorohydride/ygopro-core] Make 2nd parameter inGetLocationCountFromEx optional (#262)



It wasn't even really used in scripts, and the only times it was, tp was set as parameter, even of that isn't really the same as reason player on some cases, thus adding also the possibility of wrong interactions in the long run. Since there's no way to retrieve the reason effect (and even if there was it would have been kinda wasting to call a function to pass it as default parameter) it's better to make it optional
 
 
You can view, comment on, or merge this pull request online at:
 
  https://github.com/Fluorohydride/ygopro-core/pull/262
 
Commit Summary
    
Make 2nd parameter in GetLocationCoubtFromEx optional
  
File Changes
    
     M     libduel.cpp     (2)   
  
Patch Links:
    
https://github.com/Fluorohydride/ygopro-core/pull/262.patch
   
https://github.com/Fluorohydride/ygopro-core/pull/262.diff
  
&mdash;

edo9300:
Tp and reason player might be different values

mercury233:
`Duel.GetLocationCountFromEx(tp)` works now, so I can't get what you mean

edo9300:
Whenever it's being used with the other additional parameters such as mg or scard, tp is used in place of reason player as 2nd parameter

mercury233:
so why do you change it to nil

edo9300:
So that the core uses reason player as intended

edo9300:
So, if then using to as both values is the same, then there's no point for this 2nd parameter at all, and it could be even removed

PR #274, opened 2020-01-12
update Duel.GetCustomActivityCount

mercury233:
improve https://github.com/Fluorohydride/ygopro-scripts/pull/1306


DailyShana:
we should keep api style consistent with others, like
https://github.com/Fluorohydride/ygopro-core/blob/5dea59cb15665357949795bcc34a028e187c6024/libduel.cpp#L4135-L4139

PR #279, opened 2020-01-19
update Duel.CheckReleaseGroup

mercury233:
before:
```lua
	if sg:GetCount()>0 and Duel.GetMZoneCount(tp,sg)>0 then
		Duel.SetSelectedCard(sg)
		return Duel.CheckReleaseGroup(tp,nil,0,nil)
	else return false end
```
after:
```lua
	return sg:GetCount()>0 and Duel.GetMZoneCount(tp,sg)>0 and Duel.CheckReleaseGroup(tp,sg,#sg,nil)
```

### DIFF
--- a/libduel.cpp
+++ b/libduel.cpp
@@ -2479,11 +2479,17 @@ int32 scriptlib::duel_check_release_group(lua_State *L) {
 	int32 playerid = (int32)lua_tointeger(L, 1);
 	if(playerid != 0 && playerid != 1)
 		return 0;
+	uint32 fcount = (uint32)lua_tointeger(L, 3);
 	int32 use_con = FALSE;
 	if(check_param(L, PARAM_TYPE_FUNCTION, 2, TRUE)) {
 		use_con = TRUE;
 	} else if(check_param(L, PARAM_TYPE_GROUP, 2, TRUE)) {
 		group* pgroup = *(group**)lua_touserdata(L, 2);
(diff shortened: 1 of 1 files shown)


DailyShana:
it's misleading, Duel.CheckReleaseGroup(tp,sg,0,nil) means check 0 monsters from sg?

mercury233:
means can release sg and 0 other monster

DailyShana:
`Duel.CheckReleaseGroup(tp,filter,count,nil)` means is there exists `count` monster satisfying filter to release, so it is nature to expect that `Duel.CheckReleaseGroup(tp,group,count,nil)` means is there exists `count` monster in `group` to release.
It's inappropriate that a parameter has different meaning in two overloads.

mercury233:
@DailyShana it is updated to use `Duel.CheckReleaseGroup(tp,sg,#sg,nil)` now

DailyShana:
I think this way is enough
```lua
function filter(c,sg)
    return sg:IsContains(c)
end

return sg:GetCount()>0 and Duel.GetMZoneCount(tp,sg)>0 and Duel.CheckReleaseGroup(tp,filter,#sg,sg)
```

mercury233:
should it be `Duel.CheckReleaseGroup(tp,filter,#sg,nil,sg)` ?

DailyShana:
yes, it's a little mistake

PR #280, opened 2020-01-20
fix control adjust

DailyShana:
~There is a defect: if zones are not enough, the destruction is before swapping control.
This is because in some cases, the cards going to be destroyed occupying zones makes no zones to swap control.~

fix https://github.com/Fluorohydride/ygopro/issues/2238


DailyShana:
The correct rule is this: the turn player take control or return control, if no enough zone, destroy them, then, the other player do the same thing.
there is no swapping control.

PR #285, opened 2020-01-31
fix compatibility

mercury233:


DailyShana:
please use compiler that support c++14 properly
https://github.com/Fluorohydride/ygopro-core/blob/4cc1b18511cd1ad81486a29a51780187c8d82ce9/premake4.lua#L7

mercury233:
we should break compatibility just because some syntactic sugar

DailyShana:
it do not break c++14 compatibility

DailyShana:
It's 2020 now, all compiler stably support this standard that published 6 years ago. If you do not have compiler support c++14, download and install it.
It is impossible let everybody waiting for you

DailyShana:
sorry, it's a c++17 library feature, I didn't notice that because https://en.cppreference.com/w/cpp/utility/tuple/tuple only say it is applicable after N4387 but do not say anything about c++17

purerosefallen:
In CentOS below 7.5, the default gcc version installed by `yum` package manager is 4.x, which does not support c++14.
In addition, there are no ways to install gcc above 5 except by compiling from source. Compilling gcc from source is really a waste of time, and may by overwritten during the next `yum update`.

PR #287, opened 2020-02-01
fix disable shuffle check

DailyShana:
fix problem reported on https://github.com/edo9300/ygopro-core/issues/99


purerosefallen:
How is this one now? Any problems?

DailyShana:
maybe [only for check level > 0](https://github.com/Fluorohydride/ygopro-core/pull/287/commits/675af8477a3e5929caff3db29135a7f81a73a73b)  is unnecessary
on the other hand, `core.reason_effect` and `core.reason_player` can be disturbed by special summon cost

PR #292, opened 2020-02-21
add Duel.GetBattleMonster

mercury233:
Before:
```lua
	local a=Duel.GetAttacker()
	local d=Duel.GetAttackTarget()
	if a:IsControler(1-tp) then a,d=d,a end
```
After:
```lua
	local a,d=Duel.GetBattleMonster(tp)
```

----

Before:
```lua
	local tc=Duel.GetAttacker()
	if tc:IsControler(1-tp) then tc=Duel.GetAttackTarget() end
```
After:
```lua
	local tc=Duel.GetBattleMonster(tp)
```
----

Before:
```lua
	local a=Duel.GetAttacker()
	local d=Duel.GetAttackTarget()
	return (a:GetControler()==tp and a:IsSetCard(0x33) and a:IsRelateToBattle())
		or (d and d:GetControler()==tp and d:IsSetCard(0x33) and d:IsRelateToBattle())
```
After:
```lua
	local a=Duel.GetBattleMonster(tp)
	return a and a:IsSetCard(0x33) and a:IsRelateToBattle()
```

### DIFF
--- a/libduel.cpp
+++ b/libduel.cpp
@@ -2189,7 +2189,7 @@ int32 scriptlib::duel_get_attack_target(lua_State *L) {
 	interpreter::card2value(L, pcard);
 	return 1;
 }
-int32 scriptlib::duel_get_battle_monster(lua_State* L) {
+int32 scriptlib::duel_get_battle_monster(lua_State *L) {
 	check_param_count(L, 1);
 	check_param(L, PARAM_TYPE_INT, 1);
 	duel* pduel = interpreter::get_duel_info(L);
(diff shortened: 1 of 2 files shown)


DailyShana:
I think just `local a,d=Duel.GetBattleMonster(tp)` is enough, like getting linked zones.
Since ygo is a two-player game, accepting multiple player parameters is over-design.

libduel.cpp
```diff
@@ -2189,6 +2189,24 @@ int32 scriptlib::duel_get_attack_target(lua_State *L) {
 	interpreter::card2value(L, pcard);
 	return 1;
 }
+int32 scriptlib::duel_get_battle_monster(lua_State* L) {
```
DailyShana: format consistency
```suggestion
int32 scriptlib::duel_get_battle_monster(lua_State *L) {
```
also in header file below

PR #297, opened 2020-03-18
Old master rule compatibilities

edo9300:
@DailyShana since you started already working on the new changes, i don't think there's stuff left to change (i already tested the changes and seem ok), so just to help a little i added the checks for the master rules to keep the old behaviours intact.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -3579,6 +3579,8 @@ int32 card::is_capable_turn_set(uint8 playerid) {
 		return FALSE;
 	if(pduel->game_field->is_player_affected_by_effect(playerid, EFFECT_CANNOT_TURN_SET))
 		return FALSE;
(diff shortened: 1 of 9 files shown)


DailyShana:
These two changes are just the normal ruling change, which happened many times in the past.
And the changes much simplfy the rule thus the implement is simplified structurally. It's hard to split into old ones or new ones, like the new damage step rule in master rule 3, or the end phase rule change long ago.

edo9300:
Idk, I still see them as changes introduced by the master rule. For the split, this scenario doesn't seem to drastically change the structure in such a way that it prevents it to coexist, bit I kinda see that there could be possible issues in case the handling for the conditions is reworked in the core and this would make that rework more tedious as it'll have to address these "compatibility layers"

purerosefallen:
Just reworked this PR as a PR to master...

DailyShana:
old trigger effect ruling has added back.
since old rulings about special summon count is complicated (eg. the rule of rage with eyes of blue+a negated summon has been changed in 2018.06)
and the code is full of working around, this part will not add back

PR #307, opened 2020-04-14
add Auxiliary.PreloadUds

purerosefallen:
the function Auxiliary.PreloadUds would be called after the lua environment was set, but before every cards were loaded.
In this function we can create effects affecting globally. It made chances to create special duel rules with this function.
In addition, `./script/special.lua` would be read for reading special duel rule scripts. Those card packes in `expansions` directory could include this script for special rules.
In addition, I'm making this just ready for Rush Duel. I'm currently participating the project https://github.com/Yuzurisa/ygopro-rd , which is using this mechanic. Contributions are welcomed to this one.

### DIFF
--- a/interpreter.cpp
+++ b/interpreter.cpp
@@ -39,6 +39,7 @@ interpreter::interpreter(duel* pd): coroutines(256) {
 	scriptlib::open_duellib(lua_state);
 	scriptlib::open_debuglib(lua_state);
 	//extra scripts
+	load_script("./script/special.lua");
 	load_script("./script/constant.lua");
 	load_script("./script/utility.lua");
 }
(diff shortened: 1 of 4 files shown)


DailyShana:
using script to implement rules is obviously wrong direction

edo9300:
The core shouldn't load this kind of scripts in first place (even loading utility and constant is bad behaviour), it should all be up to the program using the api to load the needed scripts

purerosefallen:
I have seen communities overriding aux.Stringid loading stuff, or forcing players add a 'rule card' in every deck. Some examples are https://github.com/Hello23-Ygopro/ygopro-pm    .
However, this could run into problem when the specific card is not added. This problem could slightly improve those habits

edo9300:
But this don't justify hardcoding yet another lua file in the core, this loading should be done from the client, where you're free to load whatever script you load at startup, the core shouldn't care about scripts or whatsoever, it should only care of receiving a char buffer and loading that as a script. Hardcoding files into it is just bad and counter productive because whoever is then going to use this core will have to deal with adding useless stuff to make it load.

PR #308, opened 2020-04-16
use range-based for loop for effect_set

DailyShana:


effectset.h
```diff
@@ -63,6 +63,12 @@ struct effect_set {
 	effect*& at(int index) {
 		return container[index];
 	}
+	auto begin() {
+		return container.begin();
+	}
+	auto end() {
+		return container.end();
```
DailyShana, line 70: this is too naïve, it should return something correspond to `container[count]`

effectset.h
```diff
@@ -101,6 +107,12 @@ struct effect_set_v {
...
```
DailyShana, line 114: same as above

PR #313, opened 2020-05-21
Don't register field effect if it has EFFECT_FLAG_SPSUM_PARAM

edo9300:
Cards using a field summon procedure, being it special summon, normal or set, in combination with EFFECT_FLAG_SPSUM_PARAM, were erroneously registered as field effects and put in effects.aura_effect, because unlike the other summon procedure effects that have by default the target range to 0,0, making ``peffect->in_range(this)`` return false, an effect using EFFECT_FLAG_SPSUM_PARAM actually has those 2 values set to the extra parameters for the effect.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -1528,7 +1528,7 @@ void card::apply_field_effect() {
 	if (current.controler == PLAYER_NONE)
 		return;
 	for (auto& it : field_effect) {
-		if (it.second->in_range(this)
+		if (!it.second->is_flag(EFFECT_FLAG_SPSUM_PARAM) && it.second->in_range(this)
 				|| ((it.second->range & LOCATION_HAND) && (it.second->type & EFFECT_TYPE_TRIGGER_O) && !(it.second->code & EVENT_PHASE))) {
 			pduel->game_field->add_effect(it.second);
 		}
@@ -1541,7 +1541,7 @@ void card::cancel_field_effect() {
 	if (current.controler == PLAYER_NONE)
 		return;
 	for (auto& it : field_effect) {
-		if (it.second->in_range(this)
+		if (!it.second->is_flag(EFFECT_FLAG_SPSUM_PARAM) && it.second->in_range(this)
 				|| ((it.second->range & LOCATION_HAND) && (it.second->type & EFFECT_TYPE_TRIGGER_O) && !(it.second->code & EVENT_PHASE))) {
 			pduel->game_field->remove_effect(it.second);
 		}
@@ -1565,7 +1565,7 @@ void card::enable_field_effect(bool enabled) {
 				it.second->id = pduel->game_field->infos.field_id++;
 		}
 		for (auto& it : field_effect) {
-			if (it.second->in_range(this))
+			if (!it.second->is_flag(EFFECT_FLAG_SPSUM_PARAM) && it.second->in_range(this))
 				it.second->id = pduel->game_field->infos.field_id++;
 		}
 		if(current.location == LOCATION_SZONE) {
@@ -1573,7 +1573,7 @@ void card::enable_field_effect(bool enabled) {
 				it.second->id = pduel->game_field->infos.field_id++;
 		}
 		for (auto& it : target_effect) {
-			if (it.second->in_range(this))
+			if (!it.second->is_flag(EFFECT_FLAG_SPSUM_PARAM) && it.second->in_range(this))
 				it.second->id = pduel->game_field->infos.field_id++;
 		}
 		if (get_status(STATUS_DISABLED))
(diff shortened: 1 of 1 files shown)


DailyShana:
`peffect->in_range(this)` has nothing to do with target range.
Any example that has wrong behavior?

edo9300:
Ok, being in sort of hurry i realized i left out part of the issue, basically, the issue isnt' exactly ``in_range(this)``, but it's still related to that. the real issue is in ``effect::is_target``. If that ``peffect->in_range(this)`` returns true, and gets added among the other aura effects, whenever ``effect::is_target`` will be called, at the moment it arrives here https://github.com/Fluorohydride/ygopro-core/blob/master/effect.cpp#L466, if the card being checked for is in a location that matches the position for spsummon parameter, this won't return false and ``effect::is_fit_target_function`` will be called, thus calling the target in the spsummon effect with ``(e,tp)`` instead of ``(e,tp,eg,ep,ev,re,r,rp,chk,c)``, like if it was checking for a valid target for the field effect. Thus by adding that check in apply_field_effect and add_effect, that field efect will never be put among the aura ones, and thus this scenario will never happen.

This is a test puzzle

```
Debug.ReloadFieldBegin(DUEL_PSEUDO_SHUFFLE,5)
Debug.SetPlayerInfo(0,8000,0,0)
Debug.SetPlayerInfo(1,8000,0,0)

local c=Debug.AddCard(10000,0,0,LOCATION_HAND,0,POS_FACEUP,true)
e1:SetType(EFFECT_TYPE_FIELD)
e1:SetCode(EFFECT_SPSUMMON_PROC)
e1:SetProperty(EFFECT_FLAG_UNCOPYABLE+EFFECT_FLAG_SPSUM_PARAM)
e1:SetRange(LOCATION_HAND)
e1:SetTargetRange(POS_FACEUP,1)
e1:SetTarget(function(...)
	local e,tp,eg,ep,ev,re,r,rp,chk,c=...
	Debug.Message("In first target function")
	Debug.Message(#{...})
	Debug.Message(tp)
	return true
end)
c:RegisterEffect(e1)
e2=e1:Clone()
e2:SetProperty(EFFECT_FLAG_UNCOPYABLE)
e2:SetTargetRange(0,0)
e2:SetTarget(function(...)
	local e,tp,eg,ep,ev,re,r,rp,chk,c=...
	Debug.Message("In second target function")
	Debug.Message(#{...})
	Debug.Message(tp)
	return true
end)
c:RegisterEffect(e2)
local c=Debug.AddCard(10000,0,0,LOCATION_MZONE,0,POS_FACEUP,true)
Debug.ReloadFieldEnd()
e3=Effect.CreateEffect(c)
e3:SetType(EFFECT_TYPE_FIELD+EFFECT_TYPE_CONTINUOUS)
e3:SetCode(EVENT_ADJUST)
e3:SetOperation(function(e)
	Debug.Message("Check 1")
	e:GetHandler():IsHasEffect(EFFECT_SPSUMMON_PROC)
	Debug.Message("Check 2")
end)
Duel.RegisterEffect(e3,0)
```

As you can see, when IsHasEffect is called by another card, and checks for EFFECT_SPSUMMON_PROC, only the target of the summon procedure with a target range will be called, whereas the other procedure won't, and it will hve 2 parameters, e and tp.

DailyShana:
if not registered to field effect, it will not usable:
https://github.com/Fluorohydride/ygopro-core/blob/12ec3a468850212bb016f1c14e89c7e546b8a76f/processor.cpp#L2120

The problem is target range is not usual sense if there is a `EFFECT_FLAG_SPSUM_PARAM`. So we should exclude it in `effect::is_target`

btw, use "Copy permalink"
image
the code will be showed directly here

edo9300:
But on the other side, that summon effect shoudln't be put in the aura ones in first place, as it serves a different purpose, so i think this change is still needed. (i'll use the permalink the next time, thanks for the tip :wink: )

DailyShana:
if not registered to field effect, it will not usable:
https://github.com/Fluorohydride/ygopro-core/blob/12ec3a468850212bb016f1c14e89c7e546b8a76f/processor.cpp#L2118-L2127
https://github.com/Fluorohydride/ygopro-core/blob/12ec3a468850212bb016f1c14e89c7e546b8a76f/field.cpp#L1310-L1320

edo9300:
Hmm, you're right, it seems that my test environment wasn't the best one, as having a card with both types of summon procedures, still made one of the two be valid, thus making them still summonable, and since in other places filter_spsummon_procedure is used (that directly iterates only field_effect) https://github.com/Fluorohydride/ygopro-core/blob/12ec3a468850212bb016f1c14e89c7e546b8a76f/card.cpp#L2610-L2611
you wold have still been able to summon the card. Would it be reasonable to use another list to save the summon procedures to differentiate them from aura effects?

DailyShana:
I file the problem to another issue

PR #319, opened 2020-06-14
update immune effects

salix5:
@DailyShana  @mercury233 
Problem
[86.zip](https://github.com/Fluorohydride/ygopro-core/files/4776515/86.zip)
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=6890&keyword=&tag=-1
エクシーズ・オーバーディレイ(taeget: No.86 H－C ロンゴミアント with 3+ materials)
Now: No. 86 will not return to Extra Deck.
Correst: No. 86 will return to Extra Deck.

Reason
The status of No.86 is not refreshed after losing its materials.
In current version, Duel.AdjustInstantly(tc) does not work since filter_disable_related_cards() does not check the continuous effects of this.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -1580,7 +1580,6 @@ void card::enable_field_effect(bool enabled) {
 			reset(RESET_DISABLE, RESET_EVENT);
 	} else
 		set_status(STATUS_EFFECT_ENABLED, FALSE);
-	filter_immune_effect();
 	if (get_status(STATUS_DISABLED | STATUS_FORBIDDEN))
 		return;
 	filter_disable_related_cards();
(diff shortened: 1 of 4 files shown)


DailyShana:
maybe there should be an availability check in:
https://github.com/Fluorohydride/ygopro-core/blob/422b199921bea8fb22f95c8cbc7c19b8f5112b16/effect.cpp#L513-L525
and `peffect->is_available()` should be removed in `card::filter_immune_effect()`:
https://github.com/Fluorohydride/ygopro-core/blob/422b199921bea8fb22f95c8cbc7c19b8f5112b16/card.cpp#L2375-L2417

salix5:
Thank you for your advice.
Now エクシーズ・オーバーディレイ can work without Duel.AdjustInstantly(tc).

DailyShana:
the change in `filter_disable_relate_cards` is unnecessary now

salix5:
Changes to card::filter_disable_related_cards() are reverted.
Thank you.

salix5:
Is there any protblem in this branch now?

salix5:
@DailyShana 
Is there any problem in this branch now?

PR #320, opened 2020-06-18
fix trap monsters disabling under duel rule<=4

edo9300:
If under mr4 or lower a trap monster is disabled (via jinzo for example), the monster won't be able to return to its previous location, and will remain stuck in the monster zone, that because the zone locking effect is being reset on RESET_TURN_SET, but the adjust operation isn't refreshing the locations after the effects are reset (like it was doing before the mr5 update). Also, almost all the steps in trap_monster_adjust can be skipped if the duel isn't mr5 as there will always be a free location due to trap monster mechanics in previous master rules.

### DIFF
--- a/operations.cpp
+++ b/operations.cpp
@@ -1365,6 +1365,10 @@ int32 field::self_destroy(uint16 step, card* ucard, int32 p) {
 int32 field::trap_monster_adjust(uint16 step) {
 	switch(step) {
 	case 0: {
+		if(core.duel_rule <= 4) {
+			core.units.begin()->step = 3;
+			return FALSE;
+		}
 		card_set* to_grave_set = new card_set;
 		core.units.begin()->ptr1 = to_grave_set;
 		return FALSE;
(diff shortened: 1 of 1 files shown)


operations.cpp
```diff
@@ -1419,14 +1423,17 @@ int32 field::trap_monster_adjust(uint16 step) {
...
+					refresh_location_info_instant();
 				move_to_field(pcard, tp, tp, LOCATION_SZONE, pcard->current.position, FALSE, 2);
 			}
 			tp = 1 - tp;
 		}
-		card_set* to_grave_set = (card_set*)core.units.begin()->ptr1;
-		if(to_grave_set->size())
-			send_to(to_grave_set, 0, REASON_RULE, PLAYER_NONE, PLAYER_NONE, LOCATION_GRAVE, 0, POS_FACEUP);
-		delete to_grave_set;
+		if(to_grave_set) {
```
DailyShana, line 1432:
```suggestion
		if(card_set* to_grave_set = (card_set*)core.units.begin()->ptr1) {
```

DailyShana:
there is a calling of `refresh_location_info_instant()` in step 1, which is in the wrong place and should be removed

edo9300:
> there is a calling of `refresh_location_info_instant()` in step 1, which is in the wrong place and should be removed

I'm thinking if under mr5 the effects should be reset in step 1, as that's where it's then checking for the free zones, and reset them in step4 only in earlier master rules

PR #321, opened 2020-06-20
fix unique check

salix5:
@DailyShana @mercury233 
Problem
[unique.zip](https://github.com/Fluorohydride/ygopro-core/files/4807459/unique.zip)
unique1.yrp
Turn 4
Now: マジカルシルクハット cannot special summon 宮廷のしきたり if another one is on the field.
Correct: マジカルシルクハット can special summon 宮廷のしきたり if another one is on the field.

unique2.yrp
Now: Face-down 炎の王 ナグルファー cannot return to field if another one is on the field.
Correct: Face-down 炎の王 ナグルファー can return to field if another one is on the field.

unique3.yrp
Now: If there is a  炎の王 ナグルファー on the filed, the player can still choose summon when activating the effect of 死皇帝の陵墓.
The summon will not work, but the player can activate its effect again.
Correct: If there is a  炎の王 ナグルファー on the filed, the player can only choose monster set.

Reason
unique1.yrp
unique2.yrp
It should check uniqueness only when monsters enter the field in face-up position.

unique3.yrp
The result of  card::is_can_be_summoned() is wrong.
Current version: 
filter_set_procedure() returns 0, but is_can_be_summoned() returns true.
filter_set_procedure() will return 0 if check_tribute() fails, so is_can_be_summoned() will ignore res=0 if  peffec is set.

Modified version: 
1. For effects like 死皇帝の陵墓 which does not need tributes, filter_summon_procedure() should ignore tribute check and return the real checking result.
That is, filter_set_procedure() returns 0 means it really cannot be summoned for effects like 死皇帝の陵墓, and is_can_be_summoned() will return false.

2. is_player_can_summon() check is moved to filter_summon_procedure().

Correct version
[unique_correct.zip](https://github.com/Fluorohydride/ygopro-core/files/4807488/unique_correct.zip)

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -2442,7 +2442,7 @@ void card::filter_disable_related_cards() {
 // -1 = this has a EFFECT_LIMIT_SUMMON_PROC, at least 1 available
 // 0 = no EFFECT_LIMIT_SUMMON_PROC, and ordinary summon ia not available
 // 1 = no EFFECT_LIMIT_SUMMON_PROC, and ordinary summon ia available
-int32 card::filter_summon_procedure(uint8 playerid, effect_set* peset, uint8 ignore_count, uint8 min_tribute, uint32 zone) {
+int32 card::filter_summon_procedure(uint8 playerid, effect_set* peset, uint8 ignore_count, uint8 min_tribute, uint32 zone, uint8 ignore_tribute, uint32 sumtype) {
 	effect_set eset;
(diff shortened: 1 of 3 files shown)


DailyShana:
too complicated, just adding unique check in `is_player_can_summon` and `is_player_can_special_summon` is enough

DailyShana:
宮廷のしきたり
`c:SetUniqueOnField(1,0,9995766,LOCATION_SZONE)`

PR #324, opened 2020-06-26
change the order of EFFECT_REFLECT_DAMAGE

salix5:
@DailyShana @mercury233 
Problem
Fluorohydride/ygopro-scripts#1434
[reflect.zip](https://github.com/Fluorohydride/ygopro-core/files/4837451/reflect.zip)
reflect.lua(Single mode)
Now: After the effect of RR－レディネス applies, the player still takes damage from EFFECT_REFLECT_DAMAGE.
Correct: After the effect of RR－レディネス applies, the player takes no damage from EFFECT_REFLECT_DAMAGE.

Reason
The system checks EFFECT_CHANGE_DAMAGE before EFFECT_REFLECT_DAMAGE, and the result of EFFECT_REFLECT_DAMAGE will be eexcute directly.

Solution
Now it will check EFFECT_REFLECT_DAMAGE before EFFECT_CHANGE_DAMAGE.

### DIFF
--- a/operations.cpp
+++ b/operations.cpp
@@ -467,12 +467,26 @@ int32 field::damage(uint16 step, effect* reason_effect, uint32 reason, uint8 rea
 				}
 			}
 		}
+		eset.clear();
+		filter_player_effect(playerid, EFFECT_REFLECT_DAMAGE, &eset);
+		for (int32 i = 0; i < eset.size(); ++i) {
+			pduel->lua->add_param(reason_effect, PARAM_TYPE_EFFECT);
+			pduel->lua->add_param(amount, PARAM_TYPE_INT);
+			pduel->lua->add_param(reason, PARAM_TYPE_INT);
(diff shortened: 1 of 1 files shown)


operations.cpp
```diff
@@ -467,32 +467,33 @@ int32 field::damage(uint16 step, effect* reason_effect, uint32 reason, uint8 rea
 				}
 			}
 		}
-		uint32 val = amount;
 		eset.clear();
-		filter_player_effect(playerid, EFFECT_CHANGE_DAMAGE, &eset);
-		for(int32 i = 0; i < eset.size(); ++i) {
+		filter_player_effect(playerid, EFFECT_REFLECT_DAMAGE, &eset);
+		for (int32 i = 0; i < eset.size(); ++i) {
```
DailyShana, line 472: extra space

salix5: fixed
Thank you.

PR #326, opened 2020-06-30
port to lua 5.4

DailyShana:
lua 5.4 is just released recently, to port to lua 5.4, there are two places to change:

- there is a new parameter for `lua_resume` and no way to backward

- the behavior of initial calling `luaL_ref(L, LUA_REGISTRYINDEX)` is changed
in `interpreter::set_duel_info`, the return value of `luaL_ref(L, LUA_REGISTRYINDEX)` is 4 instead of 3, so in `interpreter::get_duel_info`, the first line should be changed to `lua_rawgeti(L, LUA_REGISTRYINDEX, 4)`.
as the initial count of elements in `LUA_REGISTRYINDEX` is not part of api, so we should use other keys to save `duel*` in `LUA_REGISTRYINDEX`. Here I choose the address of a static varieble as key, which is unique by the nature of c/c++.

there is another issue when using lua 5.4. When calling `c:CheckActivateEffect` during execution of a coroutine (e.g executing `target`), it will cause a "C stack overflow", which do not occur in lua 5.3.
It seem that the stack overflow check is too strict, so we have to wait the upstream to solve this issue before merging this.

### DIFF
--- a/interpreter.cpp
+++ b/interpreter.cpp
@@ -550,7 +550,8 @@ int32 interpreter::call_coroutine(int32 f, uint32 param_count, uint32 * yield_va
 	}
 	push_param(rthread, true);
 	current_state = rthread;
-	int32 result = lua_resume(rthread, 0, param_count);
+	int32 res;
+	int32 result = lua_resume(rthread, 0, param_count, &res);
 	if (result == 0) {
 		coroutines.erase(f);
 		if(yield_value)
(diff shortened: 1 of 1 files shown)


edo9300:
Are there any major benefits that the change would include? (Updating from 5.2 to 5.3 was really helpful as it introduced 64 bit values, bitwise operators, and other neat useful features) but looking at the changelog for 5.4, it doesn't seem like it's bringing any new interesting feature

DailyShana:
may be to-be-closed varieble has potential usage

purerosefallen:
Even no benefits, Lua should also be updated.
In package managers like `apt` or `yum` in Debian and CentOS, the old packages would always be outdated and may be replaced by new ones in system upgrades. Lua being outdated would make YGOPro hard to install on those OSes, or some isolated environments, like Docker or Snap, would be needed to support outdated Lua versions.

edo9300:
@purerosefallen libraries can be statically linked, and in any case most of the package managers don't offer lua built as c++, that's why by default it is statically linker already, so the distribution issue isn't a problem.

edo9300:
Anyways, @DailyShana, since the changes are very minimal, going with a simple lua version check shoudl be enough
```cpp
#if    LUA_VERSION_NUM <= 503
    int32 result = lua_resume(rthread, 0, param_count);
#else
    int32 res;
    int32 result = lua_resume(rthread, 0, param_count, &res);
#endif
```
The other change needs no version check as its compatible by default on 5.3 as well

purerosefallen:
@edo9300 Installing higher Lua versions would sometimes break `yum` package manager's function. That's why I seldom install YGOPro directly on CentOS.

PR #327, opened 2020-06-30
Pass extra parameter when retrieving the value of EFFECT_CHANGE_BATTL…

edo9300:
…E_DAMAGE

This is needed mostly for Ghostrick Mansion, that card halves the battle damage only when a non ghostrick monster is inflicting it. The current script instead will halve the battle damage whenever a non ghostrick monster battles. The extra parameter, will be a boolean indicating if the handler of that effect is the card actually inflicting the damage. This will also require changing mansion's script from using a field effect to using grant effect, so that the handler will be properly set.

```

	local e5=Effect.CreateEffect(c)
	e5:SetType(EFFECT_TYPE_SINGLE)
	e5:SetCode(EFFECT_CHANGE_BATTLE_DAMAGE)
	e5:SetValue(s.damval)
	local e6=Effect.CreateEffect(c)
	e6:SetType(EFFECT_TYPE_FIELD+EFFECT_TYPE_GRANT)
	e6:SetProperty(EFFECT_FLAG_IGNORE_IMMUNE+EFFECT_FLAG_SET_AVAILABLE)
	e6:SetRange(LOCATION_FZONE)
	e6:SetTargetRange(LOCATION_MZONE,LOCATION_MZONE)
	e6:SetTarget(s.rdtg)
	e6:SetLabelObject(e5)
	c:RegisterEffect(e6)

function s.rdtg(e,c)
	return not c:IsSetCard(0x8d)
end
function s.damval(e,damp,isres)
	if isres then return HALF_DAMAGE
	else return -1 end
end
```

### DIFF
--- a/processor.cpp
+++ b/processor.cpp
@@ -3433,9 +3433,12 @@ void field::calculate_battle_damage(effect** pdamchange, card** preason_card, ui
 							int32 val = -1;
 							if(!eset[i]->is_flag(EFFECT_FLAG_PLAYER_TARGET)) {
 								pduel->lua->add_param(p, PARAM_TYPE_INT);
+								pduel->lua->add_param(eset[i]->get_handler() == core.attacker, PARAM_TYPE_BOOLEAN);
+								val = eset[i]->get_value(2);
+							} else if(eset[i]->is_target_player(p)) {
+								pduel->lua->add_param(eset[i]->get_handler() == core.attacker, PARAM_TYPE_BOOLEAN);
 								val = eset[i]->get_value(1);
-							} else if(eset[i]->is_target_player(p))
-								val = eset[i]->get_value();
+							}
 							if(val == 0) {
 								dam_value = 0;
 								break;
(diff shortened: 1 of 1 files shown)


DailyShana:
inflicting battle damage includes the attack target in atk/def position has higher atk/def than the atk of attacker
>Odd-Eyes Pendulum Dragon
If this card battles an opponent's monster, any battle damage this card inflicts to your opponent is doubled.


also, if you want to write such effect in other place, just adding a condition is enough
```lua
local e5=Effect.CreateEffect(c)
e5:SetType(EFFECT_TYPE_SINGLE)
e5:SetCode(EFFECT_CHANGE_BATTLE_DAMAGE)
e5:SetCondition(s.damcon)
e5:SetValue(HALF_DAMAGE)

function s.damcon(e)
	return Duel.GetAttacker()==e:GetHandler()
end
```

edo9300:
@DailyShana that's wrong, it won't be applied if the monster is attacked and the attacking monster has an atk lower than the def/atk of odd eyes pendulum dragon (depending in its position), as you'll be checking for the odd eyes to be the attacking monster, which is wrong

DailyShana:
>■「オッドアイズ・ペンデュラム・ドラゴン」のモンスター効果は、自身の攻撃によって相手に戦闘ダメージを与える場合だけでなく、守備表示の「オッドアイズ・ペンデュラム・ドラゴン」が自身の守備力よりも低い攻撃力の相手モンスターに攻撃された際に相手に戦闘ダメージを与える場合などにも適用されます。
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=11213&request_locale=ja

edo9300:
Have an odd-eyes pendulum dragon in attack position, have it be attacked by a Speedroid Hexasaucer (with more than 2500 atk), the opponent will take doubled battle damage, even if it wasn't odd-eyes pendulum dragon dealing it, with this extra parameter, the odd eyes will check if it's the one actually dealing the battle damage, and since it will be the speedroid, the damage won't be changed, and this should be the proper implementation.

edo9300:
Instead, regarding the ghostrick situation, while mansion is active, attack a non ghostrick monster with a ghostrick monster, with the current implementation, the damage will be halved, when it shouldn't

DailyShana:
your interpretation of the ruling need faq

DailyShana:
Thank you, I didn't notice the difference of "battle damage it inflicts" and "battle damage from the battle it involved"

push `reason_card` as parameter for the value function directly will avoid using grand effect:
```lua
function s.damval(e,damp,rc)
    if not rc:IsSetCard(0x8d) then
        return HALF_DAMAGE
    else return -1 end
end
```

edo9300:
Yeah, passing reason card is actually better _I always tend to overcomplicate stuff_

edo9300:
Ok, now it's passing reason card

PR #333, opened 2020-07-06
Reset extra summon parameters in all the instances where card::is_spe…

edo9300:
…cial_summonable returns false

Currently if a card checks for a monster to be procedure summoned using specific cards, or a specific count, if the check returns false before checking filter_spsummon_procedure, those extra parameters aren't reset, and will be then passed the next time those functions are called.


DailyShana:
such side effect in a function makes hard to understand the code.
the reset in `card::is_special_summonable` should be removed.

if there are some place have code like this:
```
core.limit_tuner = tuner;
return pcard->is_special_summonable(p, SUMMON_TYPE_SYNCHRO);
```
change to
```
core.limit_tuner = tuner;
int32 res = pcard->is_special_summonable(p, SUMMON_TYPE_SYNCHRO);
core.limit_tuner = 0;
return res;
```

PR #334, opened 2020-08-06
fix send_to by REASON_MATERIAL

mercury233:
fix https://github.com/Fluorohydride/ygopro-scripts/commit/3e3d26d80e6da0f5952d40843d9ba631bf6fc26f
Fusion and ritual effect had checked immune in script.


DailyShana:
Because Fusion and Ritual are summon by effects, they check immune of materials.
Because Synchro, Xyz, Link are not effects, they don't check immune of materials.
The point is whether the summon method is effect, not the reason of sending to grave.

PR #336, opened 2020-09-04
Don't let trigger effects trigger if their condition is met at the same time as them changing location

edo9300:
This aims to be a proper solution for the workarounds used [here](https://github.com/Fluorohydride/ygopro-scripts/pull/1481) and [here](https://github.com/Fluorohydride/ygopro-scripts/pull/1486).
When a card is moved, it's put in the just_sent_cards set, if this happens during a chain, and in the activation validity check, it'll return false if the card is present in this set. This requires no script change, as it will all be handled internally.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -1508,6 +1508,8 @@ void card::xyz_remove(card* mat) {
 	if(mat->overlay_target != this)
 		return;
 	xyz_materials.erase(xyz_materials.begin() + mat->current.sequence);
+	if(pduel->game_field->core.current_chain.size() > 0)
+		pduel->game_field->core.just_sent_cards.insert(mat);
 	mat->previous.controler = mat->current.controler;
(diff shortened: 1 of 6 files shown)


mercury233:
I doubt it will prevent some activations...

edo9300:
IMO it's a good enough assumption, triggers in general require a location to activate in, and we have the rulings that state that the card has to be in that location at the time that the event happens, also I couldn't find cards that might not work with this, if you can find any, that might be useful.

mercury233:
Actually in my test it didn't fix the basic _Virtual World Hime - Nyannyan_ problem. Send Nyannyan to grave with Lulu, Nyannyan still be able to activate (if the `(se==nil or c:GetReasonEffect()~=se)` part in script is deleted)

mercury233:
see above comment

PR #348, opened 2020-11-25
init group.it

mercury233:
prevent crash when calling `GetNext` without `GetFirst`

### DIFF
--- a/group.cpp
+++ b/group.cpp
@@ -13,15 +13,18 @@ group::group(duel* pd) {
 	ref_handle = 0;
 	pduel = pd;
 	is_readonly = FALSE;
+	it = container.begin();
 }
 group::group(duel* pd, card* pcard) {
 	container.insert(pcard);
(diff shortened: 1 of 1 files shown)


edo9300:
The proper handling should be to check if the iterator is valid (thus if getfirst has been called already), as it makes sense to have it initialized only when you're actually iterating it. If a script is calling getnext on a group that hasn't called getfirst, then it's doing something wrong, and in case a lua error should be raised.

mercury233:
In C++ it is hard to check whether the value is set.
There are many places in ocgcore where set the value of this iterator, which increase the difficulty.

edo9300:
I know that, but it's also wrong changing that behaviour

salix5 (approved):

salix5:
My opinion:
The property "it" is a specific iterator related to this group, which always points to the current access position.
It should point to begin by default.

PR #359, opened 2021-02-24
fix update unique self destroy

mercury233:


shadowfox87:
This fixes Case 5 listed here: https://github.com/Fluorohydride/ygopro/issues/2321

PR #361, opened 2021-03-04
update Duel.GetFusionMaterial

mercury233:
for [_魔道騎竜カース・オブ・ドラゴン_](https://github.com/Fluorohydride/ygopro-pre-script/blob/6fa7146b17b47ff916923dbb89b914d66e95327c/scripts/VX01-JP/c100277001.lua)


chronogenexx:
it would be better to add LOCATION_SZONE due to Performapal Odd-Eyes Dissolver.

PR #365, opened 2021-03-10
Refresh disable status for spell activated from hand 

salix5:
@mercury233 
Problem
王宮の勅命 is on the field.

Activating a spell card from the field:
The spell card will have STATUS_DISABLED immediately.

Activating a spell card from the hand:
The spell card will not have STATUS_DISABLED immediately.

It may cause the same card has different behaviors when it is activated from different place.

Reason
The disable status is not refreshed when the spell card is moved to the field from the hand.

Solution
Refresh it in step 1.

Test replay
[replay.zip](https://github.com/Fluorohydride/ygopro-core/files/6117863/replay.zip)

### DIFF
--- a/field.cpp
+++ b/field.cpp
@@ -328,8 +328,7 @@ void field::move_card(uint8 playerid, card* pcard, uint8 location, uint8 sequenc
 					return;
 				} else
 					remove_card(pcard);
-			}
-			else if(location & LOCATION_ONFIELD) {
+			} else if(location & LOCATION_ONFIELD) {
 				if (playerid == preplayer && sequence == presequence)
 					return;
 				if(location == LOCATION_MZONE) {
@@ -373,13 +372,11 @@ void field::move_card(uint8 playerid, card* pcard, uint8 location, uint8 sequenc
 						pcard->unique_fieldid = UINT_MAX;
 				}
 				return;
-			}
-			else if(location == LOCATION_HAND) {
+			} else if(location == LOCATION_HAND) {
 				if(preplayer == playerid)
 					return;
 				remove_card(pcard);
-			}
-			else {
+			} else {
 				if(location == LOCATION_GRAVE) {
 					if(pcard->current.sequence == player[pcard->current.controler].list_grave.size() - 1)
 						return;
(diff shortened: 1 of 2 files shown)


mercury233:
there are hundreds of `} else` in the current code, should we change the format?

mercury233:
I think we should add 
`target->refresh_disable_status();`
in `move_to_field` :
https://github.com/Fluorohydride/ygopro-core/blob/b75bb919bbf0c679379890701df084a487dbf999/operations.cpp#L4755-L4757
because `change_position` and `send_to` have that step too

salix5:
> 
> there are hundreds of `} else` in the current code, should we change the format?

Maybe?
I think it is disturbing when using tools like Notepad++ or Visual Studio, since hiding sections will have some problems.

> 
> 
> I think we should add
> `target->refresh_disable_status();`
> in `move_to_field` :
> 
> https://github.com/Fluorohydride/ygopro-core/blob/b75bb919bbf0c679379890701df084a487dbf999/operations.cpp#L4755-L4757
> 
> because `change_position` and `send_to` have that step too

`move_to_field()` is called in 
`field::get_control()`
`field::control_adjust()`
`field::trap_monster_adjust()`
` field::equip()`
`field::summon()`
`field::mset()`
`field::sset()`
`field::sset_g()`
`field::special_summon_rule()`
`field::special_summon_step()`
`field::change_position()`

Since some of these functions will update card information after `move_to_field()` is called, calling `refresh_disable_status()` in `move_to_field()` may change the behavior of all these functions.
I am not sure if it is safe to do that.

mercury233:
is there any card which is affected by this status mismatch?

salix5:
format:
Temporarily reverted.
It can be discussed later.

> 
> is there any card which is affected by this status mismatch?
Whether a card can be targeted by a disable effect is determined by STATUS_DISABLED.

Now:
Activating a spell card from the hand:
The spell card X will not grant STATUS_DISABLED upon activating.
Targeting disable effect can be chained to X.

Correct:
Activating a spell card from the hand:
Just like activating a monster effect while Skill Drain on the field, the spell card X will grant STATUS_DISABLED upon activating.
Targeting disable effect cannot be chained to X.

PR #366, opened 2021-03-14
remove COUNTER_NEED_ENABLE

salix5:
Mail in 2019/1/19

Counters with COUNTER_WITHOUT_PERMIT can be put on any monster.
Before:
A card put counters of this type on X:
The counters will not be removed when X is disabled.

A card put counters of this type on itself:
The counters will be removed when the card itself is disabled, the counters put on itself is denoted by COUNTER_NEED_ENABLE.

Now:
A card put counters of this type on itself:
The counters will not be removed when the card itself is disabled.


mercury233:
_confirming_

PR #367, opened 2021-03-18
Update spell counter

salix5:
@mercury233 

Problem
I. EFFECT_COUNTER_PERMIT should be a continuous effect
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=14130&keyword=&tag=-1
The cards with EFFECT_COUNTER_PERMIT can accept Spell Counter after it is resolved.
The player cannot do:
Chain 1: 魔導変換
Chain 2: 魔導加速
This behavior is incorrect, see ddd.

II. 魔力カウンターを置く事ができるカード(cards that can have a Spell Counter)
The term has multiple definitions.
For effects that only put a counter, it means "cards where you can actually put a counter now".
It does NOT contain cards which are disabled or reach their maximum. 

https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=22526&keyword=&tag=-1
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=14213&keyword=&tag=-1
For other effects, it means "cards with EFFECT_COUNTER_PERMIT".
Ex. 創聖魔導王 エンディミオン/Endymion, the Mighty Master of Magic
It contains cards which are disabled or reach their maximum.

III.
https://ocg-rule.readthedocs.io/zh_CN/latest/c06/2019.html#id10
8/16
裁定变更：
「念动力防卫士」的效果宣言了放置有魔力指示物的卡不能使用的场合，放置的魔力指示物全部取除，也不当作可以放置魔力指示物的卡，不能对其发动「魔力掌握」。

Solution
Card.EnableCounterPermit(Card c, int countertype[, int location])
Now the card will have a EFFECT_COUNTER_PERMIT with EFFECT_FLAG_SINGLE_RANGE.

Card.IsCanAddCounter(int countertype, int count[, int singly, int location])
card::is_can_add_counter()
Check if the player can put counters on this card.
Now Card.IsCanAddCounter() must have the argument `count`.
Usage:
The card effects which only put counters.
card::add_counter()

Card.IsCanHaveCounter(countertype)
card::is_can_have_counter()
Check if this card has a EFFECT_COUNTER_PERMIT of countertype.

For cards on the field, it must have a EFFECT_COUNTER_PERMIT in the current location.
It does not contain ダーク・ヴァルキリア/Dark Valkyria which is not in the Dual state.

For other cards, possessing a EFFECT_COUNTER_PERMIT is enough.

Usage:
The property "cards that can have a Spell Counter".
Ex. 創聖魔導王 エンディミオン/Endymion, the Mighty Master of Magic

effect::is_can_be_forbidden()
Now EFFECT_COUNTER_PERMIT and EFFECT_COUNTER_LIMIT can be forbidden.

Test replays:
[replay.zip](https://github.com/Fluorohydride/ygopro-core/files/6166289/replay.zip)
bug_spell_counter1.yrp
Chain 1: 魔導変換
Chain 2: 魔導加速
It is incorrect.

correct_spell_counter1.yrp
The Pendulum Effect of 創聖魔導王 エンディミオン/Endymion, the Mighty Master of Magic should count disabled cards.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -2181,30 +2181,21 @@ int32 card::remove_counter(uint16 countertype, uint16 count) {
 	pduel->write_buffer16(count);
 	return TRUE;
 }
+// return: the player can put a counter on this or not
 int32 card::is_can_add_counter(uint8 playerid, uint16 countertype, uint16 count, uint8 singly, uint32 loc) {
 	effect_set eset;
-	if(count > 0) {
-		if(!pduel->game_field->is_player_can_place_counter(playerid, this, countertype, count))
-			return FALSE;
-		if(!loc && (!(current.location & LOCATION_ONFIELD) || !is_position(POS_FACEUP)))
-			return FALSE;
-	}
+	if (!pduel->game_field->is_player_can_place_counter(playerid, this, countertype, count))
+		return FALSE;
+	if (!loc && (!(current.location & LOCATION_ONFIELD) || !is_position(POS_FACEUP)))
+		return FALSE;
 	uint32 check = countertype & COUNTER_WITHOUT_PERMIT;
 	if(!check) {
 		filter_effect(EFFECT_COUNTER_PERMIT + (countertype & 0xffff), &eset);
 		for(int32 i = 0; i < eset.size(); ++i) {
-			uint32 prange = eset[i]->get_value();
+			uint32 prange = eset[i]->range;
 			if(loc)
 				check = loc & prange;
-			else if(current.location & LOCATION_ONFIELD) {
-				uint32 filter = TRUE;
-				if(eset[i]->target) {
-					pduel->lua->add_param(eset[i], PARAM_TYPE_EFFECT);
-					pduel->lua->add_param(this, PARAM_TYPE_CARD);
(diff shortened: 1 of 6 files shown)


mercury233:
confirming

mercury233:
what is the difference between `is_ready` and `is_available`?
`filter_single_effect` seems never been used since 2015, should we remove it?

salix5:
> what is the difference between `is_ready` and `is_available`?

X is ready:
The handler is in-range, face-up, not in summoning (Monster Card) or the effect resolution of card activation is done(Spell/Trap Card).

X->is_available()
Check if X is ready, not negated, not forbidden.
It there is a condition, check the condition and update the effect id.

X->is_ready()
It only works for single continuous effect.
Check if X is ready.
It there is a condition, check the condition without updating effect id.

> `filter_single_effect` seems never been used since 2015, should we remove it?

I agree.

salix5:
@mercury233 

mercury233:
`is_ready` may need a better name but i have no idea

salix5:
How about  `is_enabled`?

mercury233:
what about `is_single_ready`

PR #370, opened 2021-03-26
update chain::chain_operation_sort() 

salix5:
@mercury233 
Problem
Suppose that a trigger effect X is triggered multiple times in one chain.
When select_chain() is called, there will be multiple chain object with the same effect id.

Since std::sort() is not guaranteed to be stable, we should update chain::chain_operation_sort() when effect id is tied.

Solution
Require #369 
Compare chain_id when tied.


salix5:
@mercury233 
Rebase to master branch.

PR #378, opened 2021-05-06
check lua stack before pushing

mercury233:
>When you interact with the Lua API, you are responsible for ensuring consistency. In particular, you are responsible for controlling stack overflow. You can use the function lua_checkstack to ensure that the stack has enough space for pushing new elements.
https://www.lua.org/manual/5.3/manual.html#4.2
-----
Replay: [corrupted stack.zip](https://github.com/Fluorohydride/ygopro-core/files/6434220/corrupted.stack.zip)

image

In the dueling of this replay, turn 5, the player got `[string "./script/c83764718.lua"]:21: Action is not allowed here.` for _Monster Reborn_, and he activated the first avail card in hand: _Harpie's Feather Duster_.

If you open the replay directly, you may see the player activated _Monster Reborn_ and got MSG_RETRY, but this didn't happened in the duel.

Or sometimes your ygopro will crash with some error related to memory allocating.
If it didn't happen, let us corrupt the stack.

image
Add tons of `local rg=g:Filter(Card.IsRace,nil,race)` to the loop in the script, let it create many groups.

Then you may find `c83764718.target` is called with nil as all parameters, that is why _Monster Reborn_ can't be activated so _Harpie's Feather Duster_ become the first avail card and the replay can go on.
image

-----

The reason may be some operations to the lua stack make it overflowed. So we should check the stack before pushing anything to it.

### DIFF
--- a/interpreter.cpp
+++ b/interpreter.cpp
@@ -47,6 +47,8 @@ interpreter::~interpreter() {
 }
 int32 interpreter::register_card(card *pcard) {
 	//create a card in by userdata
+	luaL_checkstack(lua_state, 1, NULL);
+	luaL_checkstack(current_state, 1, NULL);
 	card ** ppcard = (card**) lua_newuserdata(lua_state, sizeof(card*));
 	*ppcard = pcard;
 	pcard->ref_handle = luaL_ref(lua_state, LUA_REGISTRYINDEX);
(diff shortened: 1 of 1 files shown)


salix5:
confirming

salix5 (approved):

PR #379, opened 2021-05-12
add Card.GetAttributeInGrave, Card.GetRaceInGrave

chronogenexx:
for _[輪廻独断](https://yu-gi-oh.jp/news_detail.php?page=details&id=1039)_


mercury233:
what about returning current status for card which is already in grave?

PR #391, opened 2021-08-14
update RNG and uniform distribution

salix5:
@mercury233 
# Problem
1.  `duel::get_next_integer()` is not a completely uniform distribution.
The whole range of mt19937 is 2^32.
If the length of the range [0, k] is not a divisor of 2^32, the probability of getting some numbers is higher than others.
This is because `mtrandom::real()` only uses division.

2. `duel::get_next_integer()` can return a out-of-range number.
`max` of mt19937 is 2^w-1 = 2^32-1 = 0xFFFFFFFF, so `rand()` can return 1 with very very low probability.
In this case, `duel::get_next_integer()` will return `h+1`.

3. mersenne_twister_engine
C++11 has a built-in `std::mersenne_twister_engine()`.
The class `mtrandom` can be replaced by the standard `std::mt19937`.

# Solution
DUEL_OLD_REPLAY
This is a duel created by old replay files.
It will use the original almost-uniform distribution (with correction of `h+1`)

no DUEL_OLD_REPLAY flag
It will use the uniform distribution.

Random Number Generator
C++11 has a built-in std::mersenne_twister_engine()
`class mtrandom` is replaced by `class mt19937`.

class mt19937
`rand()`
mersenne_twister_engine `std::mt19937`

`get_random_integer()`
uniform_int_distribution based on `std::mt19937`
This version is suggested by the answers in Stackoverflow.

`shuffle_vector()`
Fisher-Yates shuffle based on `get_random_integer()`

Unused:
std::uniform_int_distribution
`get_random_integer()` doe not use `std::uniform_int_distribution` because this distribution is implementation-defined.
Even we use the same random number sequences, the output might be different on different compilers.
I think that it will cause some troubles when exchanging replay files.

# Reference
std::mersenne_twister_engine
https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine

How to generate the same random number sequence over multiple types of compilers and kernels with <random>?
https://stackoverflow.com/questions/11641629/generating-a-uniform-distribution-of-integers-in-c


salix5:
Because it has a lot of impact on the fairness, I try to merge this branch as soon as possible.

PR #396, opened 2021-08-31
change the default duel_rule to 5

salix5:
@mercury233 
It has been 4 years since New Master Rule came out and 1 year since Master Rule 2020 came out.
In my opinion, the default rule should be set to the current rule.


salix5:
Debug.ReloadFieldBegin () has an argument `rule`.
If a puzzle script needs to use some old rule, it can specify the rule version.

mercury233:
yeah, but the creator of the puzzle may disappeared and the user can't know how to change that

salix5:
Maybe we can wrire the detail about rule version in eiki page of ygopro.
I will do it later.

PR #397, opened 2021-08-31
clean abuse of ptr type, compiles on vs x64.

zh99998:
- removes confusing `ptr` type, use pointer for objects and int32 for numbers.
- makes it possible to build on x64 visual studio. It treat `long` type as 32-bit and not suitable for 64-bit pointer.
https://docs.microsoft.com/en-us/cpp/cpp/data-type-ranges

gframe should also change `long` to pointer for duel object. pr at https://github.com/Fluorohydride/ygopro


DailyShana:
maybe its better to use standard portable integer types in <cstdint> to replace `typedef`s in common.h
ref: https://en.cppreference.com/w/cpp/header/cstdint
it provides `intptr_t` to hold pointer

using `duel*` breaks too much api

zh99998:
yes intptr is better.
commited.

salix5:
https://github.com/Fluorohydride/ygopro-core/commit/efdd63b177e481ab428eca4415d7fdbb212f0d26
Some problem in current branch:
- void interpreter::add_param(int32 param, int32 type, bool front)
`param` will be cast into `void*`, so it should be `intptr_t`

- `lua_pushinteger()` and so on
They should match the def in Lua.

- `intptr_t `
`intptr_t` can be used as a type name directly.

zh99998:
```
void interpreter::add_param(int32 param, int32 type, bool front)
param will be cast into void*, so it should be intptr_t
```

here int32 is correct, cast to void* is wrong but can't fix easily.

`params` is using one field to save different type of lua parameters, maybe int, bool(still int) and string or object pointer.
they never save 64-bit integer.

salix5:
> 
> 
> ```
> void interpreter::add_param(int32 param, int32 type, bool front)
> param will be cast into void*, so it should be intptr_t
> ```
> 
> here int32 is correct, cast to void* is wrong but can't fix easily.
> 
> `params` is using one field to save different type of lua parameters, maybe int, bool(still int) and string or object pointer.
> they never save 64-bit integer.

2c6c058914e1a7cd88f081dd4975aff292d54ec9
It should be `int32`.
Thank you!

mercury233:
```
../ocgcore/interpreter.cpp: In member function ‘void interpreter::push_param(lua_State*, bool)’:
../ocgcore/interpreter.cpp:193:33: error: cast from ‘void*’ to ‘int32 {aka int}’ loses precision [-fpermissive]
    lua_pushboolean(L, (int32)it.first);
                                 ^~~~~
../ocgcore/interpreter.cpp:217:32: error: cast from ‘void*’ to ‘int32 {aka int}’ loses precision [-fpermissive]
    function2value(L, (int32)it.first);
                                ^~~~~
../ocgcore/interpreter.cpp:221:28: error: cast from ‘void*’ to ‘int32 {aka int}’ loses precision [-fpermissive]
    int32 index = (int32)it.first;
                            ^~~~~
```

salix5:
https://github.com/Fluorohydride/ygopro-core/commit/6d38a9a1822c79abbdef319817834d29b7337f08
Maybe we can
1. `void*` to `intptr_t` (cast to integer with the same length)
2.  `intptr_t` to `int32` (integer conversion)

common.h
```diff
@@ -8,13 +8,12 @@
...
+#include <cstdint>
 typedef unsigned long long uint64;
 typedef unsigned int uint32;
 typedef unsigned short uint16;
 typedef unsigned char uint8;
 typedef unsigned char byte;
-typedef long ptr;
 typedef long long int64;
 typedef int int32;
 typedef short int16;
```
DailyShana, line 19: why not using standard integer types to redefine these unportable types?

common.h
```diff
@@ -8,13 +8,12 @@
 #ifndef COMMON_H_
 #define COMMON_H_
 
-typedef unsigned long uptr;
+#include <cstdint>
```
DailyShana, line 11: In my understanding, common.h is used in ocgapi.h which is a C header (because it exports C api), so here should include stdint.h

ocgapi.h
```diff
@@ -34,22 +35,22 @@ byte* read_script(const char* script_name, int* len);
...
-extern "C" DECL_DLLEXPORT void new_tag_card(ptr pduel, uint32 code, uint8 owner, uint8 location);
-extern "C" DECL_DLLEXPORT int32 query_card(ptr pduel, uint8 playerid, uint8 location, uint8 sequence, int32 query_flag, byte* buf, int32 use_cache);
-extern "C" DECL_DLLEXPORT int32 query_field_count(ptr pduel, uint8 playerid, uint8 location);
-extern "C" DECL_DLLEXPORT int32 query_field_card(ptr pduel, uint8 playerid, uint8 location, int32 query_flag, byte* buf, int32 use_cache);
-extern "C" DECL_DLLEXPORT int32 query_field_info(ptr pduel, byte* buf);
-extern "C" DECL_DLLEXPORT void set_responsei(ptr pduel, int32 value);
-extern "C" DECL_DLLEXPORT void set_responseb(ptr pduel, byte* buf);
-extern "C" DECL_DLLEXPORT int32 preload_script(ptr pduel, const char* script, int32 len);
+extern "C" DECL_DLLEXPORT intptr_t create_duel(uint32 seed);
+extern "C" DECL_DLLEXPORT void start_duel(intptr_t pduel, int32 options);
```
DailyShana, line 39: Don't you think it's strange to mix standard `intptr_t` and nonstandard `int32`？

common.h
zh99998: because the `ptr` type is designed to use as both int32 and pointer.

ocgapi.h
zh99998: should we replace all int32 to standard int ?

zh99998: and other non-standard type to standard.

common.h
DailyShana: but you have changed the int32 usage of `ptr` to `int32`, now just treat `ptr` as pointer.

ocgapi.h
DailyShana: I think replacing nonstandard types to standard is the best.
But others may think the change is too big

common.h
salix5: I don't quite understand.
If `ocgapi.cpp` is written in C++, we should use the C++ version `<cstdint>`.
Is it correct?

DailyShana: If a C program wants to import ocgcore.dll, it only needs ocgapi.h to import the symbols, so all content of ocgapi.h should be pure C.
And stdint.h is also available in C++,

salix5: fixed
Thank you!

ocgapi.h
salix5: > 
> 
> Don't you think it's strange to mix standard `intptr_t` and nonstandard `int32`？

In my ononipn, `intptr_t` is not a basic type and should be written explicitly.
So it is not so strange to use `intptr_t` and `int32` at the same time.

DailyShana: although not a basic type, `intptr_t` is still defined by standard, its behavior is definite
but our nonstandard int8/16/32/64... is still possiblly not portable

changing int32 to int32_t is the best, then we can alway working on standard integer types
another way is to redefine our nonstandard types like
```
typedef int32_t int32
typedef intptr_t ptr
...
```
keeping signature of ocgapi "unchanged"

salix5:
？

purerosefallen:
Time to reconsider this. we may use `f89f74c` as the commit

purerosefallen:
fixed the version tree of this PR. Latest one is also being merged.

purerosefallen:
This PR is not really merged, but somehow that day I pushed the wrong branch here, resetted.
Use https://github.com/Fluorohydride/ygopro-core/pull/451 this PR instead.

PR #398, opened 2021-09-08
clear core.limit_syn etc. on summon fail

mercury233:
[2021-09-06 12-58-31 Knight of Hanoi VS Knight of Hanoi.zip](https://github.com/Fluorohydride/ygopro-core/files/7127647/2021-09-06.12-58-31.Knight.of.Hanoi.VS.Knight.of.Hanoi.zip)
Turn 6, the synchro effect of _Shooting Riser Dragon_ failed, but `core.limit_tuner` was not reset, and caused the following link summon failed.


salix5:
confirming

salix5:
I prefer the solution in #399.
In my opinion, the real problem is that we should not use global variables to pass the arguments.

PR #399, opened 2021-09-10
add struct material_info

salix5:
# Problem
class card
is_spsummonable(), is_special_summonable(), filter_spsummon_procedure()
They need material info as arguments, but it is passed to those functions by `core.limit_*` now.
Passing the arguments by global variables might cause lots of problems.

# Solution
Now they will use arguments just like normal functions.


salix5:
@mercury233 

PR #416, opened 2021-11-23
push self_table and self_code to lua for GetID

mercury233:
```lua
function GetID()
	local offset=self_code<100000000 and 1 or 100
	return self_table,self_code,offset
end
```

copied from https://github.com/edo9300/ygopro-core/blob/master/interpreter.cpp#L154


salix5 (approved):

salix5:
Thank you!

purerosefallen:
How about disable debug lib as well?

PR #437, opened 2022-03-05
update shuffle_vector()

salix5:
@mercury233 @edo9300 
https://github.com/Fluorohydride/ygopro-scripts/pull/1820
3fe5cbdd991b13ab554f573a03787a13a24a7423
It should be shuffling `svector[0]` ~ `svector[s-1]`.
Is this correct?

### DIFF
--- a/field.cpp
+++ b/field.cpp
@@ -957,14 +957,14 @@ void field::shuffle(uint8 playerid, uint8 location) {
 		}
 	}
 	if(location == LOCATION_HAND || !(core.duel_options & DUEL_PSEUDO_SHUFFLE)) {
-		uint32 s = (uint32)svector.size();
+		int32 s = (int32)svector.size();
 		if(location == LOCATION_EXTRA)
-			s = s - player[playerid].extra_p_count;
+			s = s - (int32)player[playerid].extra_p_count;
 		if(s > 1) {
 			if (core.duel_options & DUEL_OLD_REPLAY)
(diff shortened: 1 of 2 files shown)


purerosefallen:
I suggest keeping the method same on `_old`, but just adding stub params.
The duels after 1.035.3 have no chance using the old shuffle, while the old things should be kept in old way, for best possible compatibility on old replays.

purerosefallen:
I suggest keeping the method same on _old, but just adding stub params.
The duels after 1.035.3 have no chance using the old shuffle, while the old things should be kept in old way, for best possible compatibility on old replays.

edo9300:
@purerosefallen most the "old" replays would have the proper behaviour in any case, as this change was there for less than the correct implementation. ~~I'll leave this up to you tho, in any csse with the normal yrp format doesn't even really make sense to keep the "old" version at all~~

salix5:
> I suggest keeping the method same on `_old`, but just adding stub params.
I don't understand this one.
Can you explain it explicitly?
Thank you.

salix5:
@mercury233 
Can we merge this pull request?

mercury233:
_too lazy to test them_

salix5:
[shuffle_test1.zip](https://github.com/Fluorohydride/ygopro-core/files/8197776/shuffle_test1.zip)
Turn 3
Extra: 5 face-up, 2 face-down

PR #450, opened 2022-06-12
Update Card.IsRelateToEffect

Yuzurisa:
The effect relation check should work in current chain resolving only, rather than any possible chains.
This solves https://github.com/Fluorohydride/ygopro/issues/2389 , without changing any scripts.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -2073,15 +2073,15 @@ void card::create_relation(effect* peffect) {
 }
 int32 card::is_has_relation(effect* peffect, uint8 any_chain) {
 	if(!any_chain) {
-			for(auto cit = pduel->game_field->core.current_chain.rbegin(); cit != pduel->game_field->core.current_chain.rend(); ++cit) {
+		for(auto cit = pduel->game_field->core.current_chain.rbegin(); cit != pduel->game_field->core.current_chain.rend(); ++cit) {
 			if(peffect == cit->triggering_effect && relate_effect.find(std::make_pair(peffect, cit->chain_id)) != relate_effect.end())
 				return TRUE;
 		}
 	} else {
-			for(auto& it : relate_effect) {
-				if(it.first == peffect)
-					return TRUE;
-			}
+		for(auto& it : relate_effect) {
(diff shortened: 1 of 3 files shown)


mercury233:
it will break this kind of check in condition and target

Yuzurisa:
In fact we have to tell what exactly a relation to an *effect* is.
Relations are defined to establish between a card and a *chain*, rather than an *effect*. Mostly we used `Card.IsRelateToEffect` in effect operations to check if a card is related to a specific effect activated  in any *current* chain, rather than any chains *to be triggered*.  In fact the reason why https://github.com/Fluorohydride/ygopro/issues/2389 occurred was, we in the past got the definition of `Card.IsRelateToEffect` mixed together.

> it will break this kind of check in condition and target

Those checks are actually another meaning of `Card.IsRelateToEffect` that we don't often use, which is to check whether a relation exists between a specific effect and any chains either resolving or to be triggered. In such cases, we may either add another param to `Card.IsRelateToEffect`, or create another function like `Card.IsRelateToAnyEffect`, to separate the definitions of those two different use.

Yuzurisa:
What's more, I have checked every use of `Card.IsRelateToEffect` in `ygopro-scripts` inside condition or targets, while all of them are in EVENT_CHAINING, EVENT_CHAIN_SOLVING, or EVENT_CHAIN_SOLVED checks, in which relations are formed already. So it wouldn't break when checking conditions or targets.
I also tested cards such as *Confronting the "C"*, in which `Card.IsRelateToEffect` is used in continuous checks. They also worked well, without using the 3rd param.

edo9300:
The issue here lies in the fact that the Effect object (which is unique and shared, like a Singleton) is being used all over the place to store individual properties tied to other factors, from the card relation which should be tied to the chain instead, the labels that could easily overlap if the same effect is used multiple times in the same chain, and other things

Yuzurisa:
@edo9300 in such cases checking relations to chain, use `Card.IsRelatedToChain` instead.

PR #454, opened 2022-07-06
add Effect.IsCostChecked

mercury233:
Some cards have cost releated to target (_Transmodify_, _Grinning Grave Virus_), but sometime those cards may be used without cost (_Destiny HERO - Diamond Dude_, _Junk Collector_).

Now we are using those codes to make sure the cost is checked:
```lua
function s.cost(e,tp,eg,ep,ev,re,r,rp,chk)
	e:SetLabel(100)
	return true
end
function s.target(e,tp,eg,ep,ev,re,r,rp,chk)
	if chk==0 then
		if e:GetLabel()~=100 then return false end
		e:SetLabel(0)
		return ......
	end
	e:SetLabel(0)
	......
end
```

And it can be shorten to `e:IsCostChecked()`

### DIFF
--- a/effect.cpp
+++ b/effect.cpp
@@ -63,6 +63,7 @@ effect::effect(duel* pd) {
 	target = 0;
 	value = 0;
 	operation = 0;
+	cost_checked = FALSE;
 }
 int32 effect::is_disable_related() {
 	if (code == EFFECT_IMMUNE_EFFECT || code == EFFECT_DISABLE || code == EFFECT_CANNOT_DISABLE || code == EFFECT_FORBIDDEN)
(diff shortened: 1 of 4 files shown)


Yuzurisa:
An example of how to write them in the new way?

mercury233:
```lua
function s.target(e,tp,eg,ep,ev,re,r,rp,chk)
	if chk==0 then
		return e:IsCostChecked() and ......
	end
	......
end
```
empty `s.cost` is no longer needed

chronogenexx:
Junk Collector and Trap of Darkness can select normal trap that used `Effect.IsCostChecked`.

Lyris12:
> Junk Collector and Trap of Darkness can select normal trap that used `Effect.IsCostChecked`.

Same with Labrynth Barrage?

mercury233:
@chronogenexx is there still any problem with `Effect.IsCostChecked` ?

chronogenexx:
https://github.com/Fluorohydride/ygopro-core/pull/460 this commit was merged.
so I think there is no problem.

mercury233:
@chronogenexx Thank you!

PR #456, opened 2022-07-20
add Effect.GetActiveCode, Effect.IsActiveCode, Effect.IsActiveSetCard

465uytrewq:
update: add `Effect.GetActiveCode`, `Effect.IsActiveCode` and `Effect.IsActiveSetCard`.
e.g.
```lua
function c59297550.spcon(e,tp,eg,ep,ev,re,r,rp)
	return re:IsActiveType(TYPE_MONSTER)
		and re:IsActiveSetCard(0x58) and not re:IsActiveCode(59297550)
end
```

ref: https://github.com/Fluorohydride/ygopro-scripts/pull/1171


salix5:
The setcode will be changed into std vector, and some parts might have to be changed.
I will try to create the PR today.

465uytrewq:
I confirmed and tested https://github.com/Fluorohydride/ygopro-core/pull/532.
I will update this commit after that commit is merged.

PR #464, opened 2022-08-30
fix card::add_effect()

salix5:
# Problem
#463
The changes in #408 will reject some normal effects with bit 0xf000.

# Reason
The effect types have 2 categories:

A. trigger-like
```cpp
EFFECT_TYPE_ACTIVATE
EFFECT_TYPE_TRIGGER_O
EFFECT_TYPE_TRIGGER_F
EFFECT_TYPE_QUICK_O
EFFECT_TYPE_QUICK_F
```
The code is an event.

B. other
The code is an effect code.

# Solution
Now the check is limited to trigger-like effects.
These effects cannot use continuous_event.


Aephiex:
This should be enough to fix it.

PR #470, opened 2022-09-07
Remove unused Duel.GetEnvironment

Aephiex:
In consistance with https://github.com/Fluorohydride/ygopro-core/pull/425

@mercury233


mercury233:
I prefer to remove this function because it is not used, at least not used after MR3 allowed dual field.

Aephiex:
Najiu remove it?

Aephiex:
@mercury233 
Duel.GetEnvironment removed.

PR #477, opened 2022-09-28
add premake file to build ocgcore.dll with lua static linked

mercury233:


.gitignore
```diff
@@ -0,0 +1,3 @@
+/premake5.exe
```
purerosefallen, line 1: Make it to `/premake5*` so that non-Windows users could build

mercury233: I did't test on other systems, and don't know why the non-Windows users need `dll` file.

PR #479, opened 2022-11-06
update Duel.Overlay

mercury233:
Requires https://github.com/Fluorohydride/ygopro/pull/2468

**Problem 1:**
All operations in YGOPro which make Xyz monster leave the field will handle xyz materials automatically,
but `Duel.Overlay` won't, you must do this in script, it is strange since if you want to overlay the xyz materials to the new monster (effect of _Number 75: Bamboozling Gossip Shadow_) you should use `Duel.Overlay` to the xyz materials ahead.
```lua
local og=tc:GetOverlayGroup()
if og:GetCount()>0 then
	Duel.SendtoGrave(og,REASON_RULE)
end
Duel.Overlay(c,Group.FromCards(tc))
```

**Problem 2:**
If we change the order of `Duel.SendtoGrave` and `Duel.Overlay` in the above script, the YGOPro client will crash, because it can't find the overlay handler to deattach from.

**Problem 3:**
If _Number 101: Silent Honor ARK_ select _Kashtira Arise-Heart_ to attach, the xyz materials of Arise-Heart should be sent to grave after Arise-Heart is attached, so them shouldn't be redirected to banish.
But now in YGOPro them will be banished because `Duel.SendtoGrave` is called before `Duel.Overlay`.

### DIFF
--- a/libduel.cpp
+++ b/libduel.cpp
@@ -3420,12 +3420,17 @@ int32 scriptlib::duel_overlay(lua_State *L) {
 		pgroup = *(group**) lua_touserdata(L, 2);
 	} else
 		return luaL_error(L, "Parameter %d should be \"Card\" or \"Group\".", 2);
-	if(pcard) {
-		card::card_set cset;
+	card::card_set cset;
+	if(pcard)
 		cset.insert(pcard);
-		target->xyz_overlay(&cset);
-	} else
-		target->xyz_overlay(&pgroup->container);
+	else
+		cset.insert(pgroup->container.begin(), pgroup->container.end());
+	card::card_set oset;
+	for(auto& pcard : cset)
+		oset.insert(pcard->xyz_materials.begin(), pcard->xyz_materials.end());
+	target->xyz_overlay(&cset);
+	duel* pduel = interpreter::get_duel_info(L);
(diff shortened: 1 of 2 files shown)


libduel.cpp
```diff
@@ -3420,12 +3420,17 @@ int32 scriptlib::duel_overlay(lua_State *L) {
...
-	} else
-		target->xyz_overlay(&pgroup->container);
+	else
+		cset.insert(pgroup->container.begin(), pgroup->container.end());
+	card::card_set oset;
+	for(auto& pcard : cset)
+		oset.insert(pcard->xyz_materials.begin(), pcard->xyz_materials.end());
+	target->xyz_overlay(&cset);
+	duel* pduel = interpreter::get_duel_info(L);
+	pduel->game_field->send_to(&oset, pduel->game_field->core.reason_effect, REASON_RULE, pduel->game_field->core.reason_player, PLAYER_NONE, LOCATION_GRAVE, 0, POS_FACEUP);
```
chronogenexx, line 3433: why not you don't use `REASON_LOST_OVERLAY`?
https://github.com/Fluorohydride/ygopro-core/blob/master/operations.cpp#L4613
these code is used `REASON_LOST_OVERLAY`.

purerosefallen:
still useful?

PR #493, opened 2023-04-15
update sum_param of MSG_SELECT_SUM

mercury233:
fix https://github.com/Fluorohydride/ygopro-scripts/pull/1785

**Problem:** 
Some cards have 2 levels when using them as material, such as _Tuningware_.
The implement of `Group.CheckWithSum...` or `Group.SelectWithSum...` function use one `unsigned int32` to handle this, use 16 bit of this param to store the normal level, and another 16 bit to store the another level:
https://github.com/Fluorohydride/ygopro-scripts/blob/master/c92676637.lua#L25
https://github.com/Fluorohydride/ygopro-core/blob/bf2339ce96152b6825c5a4064ac669def6caf976/field.cpp#L2871-L2872

This will work for level because as most times the level of one monster won't reach the limit of 16 bit, which is 65535, let alone when using as material.
But attack & defense may go beyond the limit.

Solution:
No monster need 2 attack values to calc sum. So we can use the first bit of `opParam` as mark, if it is 1, treat the other 31 bit of 
`opParam` as the value.


salix5:
confirming
maybe need a little more time

playerop.cpp
```diff
@@ -605,11 +605,19 @@ int32 field::select_counter(uint16 step, uint8 playerid, uint16 countertype, uin
...
+		op2 = 0;
+	}
+}
+static int32 select_sum_check1(const uint32* oparam, int32 size, int32 index, int32 acc, int32 opmin) {
 	if(acc == 0 || index == size)
 		return FALSE;
-	int32 o1 = oparam[index] & 0xffff;
-	int32 o2 = oparam[index] >> 16;
+	int32 o1, o2;
+	get_sum_params(oparam[index], o1, o2);
```
salix5, line 620: If we use static member:
```cpp
field::get_sum_params(oparam[index], o1, o2);
```

PR #497, opened 2023-05-15
fix Duel.SwapDeckAndGrave

mercury233:
If a ritual monster is summoned and sent to grave, it will has `STATUS_PROC_COMPLETE`, but if it is returned to deck by the effect of _Exchange of the Spirit_, and sent to grave again, it shouldn't has that status.

Test puzzle: [test_SwapDeckAndGrave.zip](https://github.com/Fluorohydride/ygopro-core/files/11474929/test_SwapDeckAndGrave.zip)


salix5:
Should we reset all status?

mercury233:
I think the other STATUSes are irrelevant to that situation

PR #502, opened 2023-08-01
fix field::recover(), field::damage()

salix5:
# Problem
1. The ATK, DEF are int32, but the max of damage is 0xffff.

2. The recover may cause overflow.

# Solution
1. Now `amount` is put in `arg3`, and it is int32.

2. Now the recover is capped at 
```cpp
int32 limit = INT32_MAX - player[playerid].lp;
```
If `amount` > `limit`, the system will:
- recover `limit` LP
- raise the event `EVENT_RECOVER` with `amount`


salix5:
@mercury233 
Can we merge this one?

PR #515, opened 2023-11-11
update `EFFECT_UNRELEASABLE_NONSUM`

Lyris12:
for `Ritual Beast Ulti-Reirautari` (`聖霊獣騎 レイラウタリ`):
* expose reason effect
* add reason support for Tributing cards

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -3595,8 +3595,15 @@ int32 card::is_releasable_by_nonsummon(uint8 playerid) {
 		return FALSE;
 	if(!pduel->game_field->is_player_can_release(playerid, this))
 		return FALSE;
-	if(is_affected_by_effect(EFFECT_UNRELEASABLE_NONSUM))
-		return FALSE;
+	effect_set eset;
+	filter_effect(EFFECT_UNRELEASABLE_NONSUM, &eset);
+	for(int32 i = 0; i < eset.size(); ++i) {
+		pduel->lua->add_param(this, PARAM_TYPE_CARD);
+		pduel->lua->add_param(playerid, PARAM_TYPE_INT);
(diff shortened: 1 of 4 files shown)


salix5:
How about adding new parameter to `Duel.CheckReleaseGroup`, `Duel.CheckReleaseGroupEx`?

For example:
```lua

---@return boolean
---@param reason integer
---@param player integer
---@param f function|nil
---@param count integer
---@param ex Card|Group|nil
---@param ... any
function Duel.CheckReleaseGroup(reason,player,f,count,ex,...) end
```

Lyris12:
> How about adding new parameter to `Duel.CheckReleaseGroup`, `Duel.CheckReleaseGroupEx`?
> 
> For example:
> 
> ```lua
> ---@return boolean
> ---@param reason integer
> ---@param player integer
> ---@param f function|nil
> ---@param count integer
> ---@param ex Card|Group|nil
> ---@param ... any
> function Duel.CheckReleaseGroup(reason,player,f,count,ex,...) end
> ```

That does not cover cards using only `Duel.GetReleaseGroup` for their costs.

salix5:
Thank you!

To-do-list:
- Duel.CheckReleaseGroup
- Duel.GetReleaseGroup

Is there anything else?

salix5:
#516 
I think we can change these functions.

Lyris12:
Not sure what the `must_param` is for, but that works for me, too!

Lyris12:
..._PROVIDED_ the scripts are updated accordingly, 1st.

Lyris12:
> > Not sure what the `must_param` is for, but that works for me, too!
> 
> The number of necessary parameters.

Ah, gotcha! Thanks!

salix5:
Since it is fixed in #516, this PR will be closed.
Thank you for your help.

PR #532, opened 2024-01-11
update setcode check

salix5:
@mercury233 
@purerosefallen 
@465uytrewq 

- Change setcode into `std::vector<uint16_t>`.
Now we can set 5+ setcodes in 1 card.

- Simplify the setcode check.

## class card_data
is_setcode()
bool is_setcode(uint32 value) const
Check if the setcode conatins `value`.

bool is_alternative() const
Check if this card is an alternative artwork card.

void set_setcode(uint64 value)
Set the setcode with `value` (Probably from sqlite file).

## class card
static bool check_card_setcode(uint32 code, uint32 value)
Check if the setcode of card `code` contains `value`.


salix5:
@mercury233 
@purerosefallen 
Can we merge this one?

edo9300:
This will break any non c++ project using the core, also will make a build of the core depend on a specific abi for std::vector

DailyShana:
There are so many projects depend on the natural of C-abi of ocgcore.

Add another data field `setcode2`, then we have 4 more setcodes
although a break change, at least it's still C abi 
```
card_data {
...
int64 setcode
int64 setcode2
...
}
```

salix5:
@edo9300 
@DailyShana 
setcode is changed to C-array.
Thank you.

salix5:
@mercury233 
@purerosefallen 
@edo9300 
@DailyShana 
Is there any other problems in this branch?

PR #516, opened 2023-11-20
update release functions

salix5:
# Problem
EFFECT_CANNOT_RELEASE cannot check reason.
It is necessary for Ritual Beast Ulti-Reirautari (聖霊獣騎 レイラウタリ).

# Solution
https://github.com/Fluorohydride/ygopro-scripts/pull/2242

target function of EFFECT_CANNOT_RELEASE
Target function:
add 4th param: reason
```lua
function s.rellimit(e,c,tp,r)
	return r&REASON_COST~=0
end
```

## Lua functions
class Card
Card.IsReleasable 
add 2nd param: reason (optional, default: REASON_COST)

class Duel
CheckReleaseGroup
SelectReleaseGroup
CheckReleaseGroupEx
SelectReleaseGroupEx
add 1st param: reason

GetReleaseGroup
GetReleaseGroupCount
add 3rd param: reason (optional, default: REASON_COST)

IsPlayerCanRelease
add 3rd param: reason (optional, default: REASON_COST)

## Release Reason
Now the main reasons are:
`REASON_ACTION`: "This card cannot declare an attack unless"...
`REASON_COST`: The card is released by the cost of activation.
`REASON_EFFECT`: The card is released by effect.
`REASON_MAINTENANCE`: The maintenance cost.
`REASON_SPSUMMON`: The card is released by the Special Summon procedure.
`REASON_SUMMON`: The card is released by Advance Summon.
`REASON_RULE`: The card is is released by non-effect actions like 痛み分け/Share the Pain.

# Test
[test_cannot_release.zip](https://github.com/Fluorohydride/ygopro-core/files/13406736/test_cannot_release.zip)

test_cannot_release.lua
Single mode script
Add test card with id=1
Level 1/Earth/Warrior/ATK 1/DEF 1

c1.lua
The script of 1
EFFECT_CANNOT_RELEASE with reason

インフェルノイド・アシュメダイ
Infernoid Piaty
From https://github.com/Fluorohydride/ygopro-scripts/pull/2222

When `1` is on the field:
The 2nd effect of `Infernoid Piaty` cannot be activated.
`Black Illusion Ritual` can be activated.

PR #534, opened 2024-01-12
update CheckReleaseGroupEx and so on

mercury233:
Change the syntax of https://github.com/Fluorohydride/ygopro-core/pull/516 

Change `Duel.CheckReleaseGroup` and `Duel.SelectReleaseGroup` back to the former syntax, remove the `reason` param added in #516, make `REASON_COST` as the reason.

Change `Duel.CheckReleaseGroupEx` to
```Duel.CheckReleaseGroupEx(player,f,count,reason,use_hand,ex,...)```
Change `Duel.SelectReleaseGroupEx` to
```Duel.SelectReleaseGroupEx(player,f,min,max,reason,use_hand,ex,...)```

If the reason of release is not Cost, it should use `Duel.CheckReleaseGroupEx` , which can set reason.

Former we use `Duel.CheckReleaseGroupEx` for checking monster zone + hand, now it requires `use_hand` param to be true.

### DIFF
--- a/libduel.cpp
+++ b/libduel.cpp
@@ -2664,58 +2664,56 @@ int32 scriptlib::duel_get_release_group_count(lua_State *L) {
 	return 1;
 }
 int32 scriptlib::duel_check_release_group(lua_State *L) {
-	const int32 must_param = 5;
+	const int32 must_param = 4;
 	check_param_count(L, must_param);
-	uint32 reason = (uint32)lua_tointeger(L, 1);
-	int32 playerid = (int32)lua_tointeger(L, 2);
+	int32 playerid = (int32)lua_tointeger(L, 1);
 	if(playerid != 0 && playerid != 1)
 		return 0;
 	int32 use_con = FALSE;
-	if(!lua_isnil(L, 3)) {
-		check_param(L, PARAM_TYPE_FUNCTION, 3);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by purerosefallen** [comment]
So why changing the API over and over again?

We have to think that somehow we had problems in communications. We can't let all ygopro devs or maintainers from other related communities looking us for fun. This really destroys the reputation of we maintainers.

**Review by salix5 — APPROVED** [review]

**Comment by salix5** [comment]
For those who has updated the scrtipt before:
Use the bash commands

```bash
sed -i -e 's/Duel\.CheckReleaseGroup(REASON_COST,/Duel.CheckReleaseGroup(/g' *.lua
sed -i -e 's/Duel\.SelectReleaseGroup(REASON_COST,/Duel.SelectReleaseGroup(/g' *.lua
sed -i -e 's/Duel\.CheckReleaseGroupEx(REASON_COST,/Duel.CheckReleaseGroupEx(/g' *.lua
sed -i -e 's/Duel\.SelectReleaseGroupEx(REASON_COST,/Duel.SelectReleaseGroupEx(/g' *.lua
```

They can restore the scripts to the previous version, and they only need to change `Duel.CheckReleaseGroupEx` and `Duel.SelectReleaseGroupEx`.

PR #540, opened 2024-01-23
fix field::equip(uint16 step,...

salix5:
@mercury233 
@purerosefallen 
- Change equip target
If the equip card is TYPE_EQUIP, it does not need to change/add type.

- Union monster
The Union monster equipped by its own effect:
type: TYPE_EQUIP + TYPE_SPELL with EFFECT_UNION_STATUS
It does not need a special type.

### DIFF
--- a/operations.cpp
+++ b/operations.cpp
@@ -1404,7 +1404,7 @@ int32 field::equip(uint16 step, uint8 equip_player, card * equip_card, card * ta
 	}
 	case 1: {
 		equip_card->equip(target);
-		if(!(equip_card->data.type & TYPE_EQUIP)) {
+		if(!(equip_card->data.type & TYPE_EQUIP) && !(equip_card->get_type() & TYPE_EQUIP)) {
 			effect* peffect = pduel->new_effect();
 			peffect->owner = equip_card;
 			peffect->handler = equip_card;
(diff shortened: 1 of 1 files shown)


mercury233:
Which card will this change fix?

salix5:
Card type
We should keep the card type as simple as possible.
There are 2 types of Equip cards:
- Equip Spell
- Trap becoming Equip card

The extra type ` TYPE_EQUIP + TYPE_SPELL + TYPE_UNION`  is  unnecessary.
I have checked the script using TYPE_UNION,  and it does not affect current scripts.

Changing equip target
https://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=15968&request_locale=ja
驚楽園の大使 ＜Bufo＞
Amazement Ambassador Bufo
30829071
②：モンスターに装備されている自分の「アトラクション」罠カード１枚を対象として発動できる。そのカードを自分フィールドの「アメイズメント」モンスターまたは相手フィールドの表側表示モンスター１体に装備する。この効果は相手ターンでも発動できる。

The 2nd effect can change the equip target in each turn, and it will create an  `EFFECT_ADD_TYPE` effect every time.
This is unnecessary because the Trap card is already an Equip card.

PR #541, opened 2024-01-24
Use array to store setcode

salix5:
@mercury233 
@465uytrewq 
由於sqlite3的int最大是8 bytes
改用陣列儲存setcode以便對應5組以上的情況

- Change setcode into `uint16_t setcode[SIZE_SETCODE]`.
Now we can set 5+ setcodes in 1 card.

- Simplify the setcode check.

## class card_data
is_setcode()
bool is_setcode(uint32 value) const
Check if the setcode conatins `value`.

bool is_alternative() const
Check if this card is an alternative artwork card.

void set_setcode(uint64 value)
Set the setcode with `value` (Probably from sqlite file).

## class card
static bool check_card_setcode(uint32 code, uint32 value)
Check if the setcode of card `code` contains `value`.


salix5:
@mercury233 
@purerosefallen 
Can we merge this one?

DailyShana:
If hard coding cannot be avoided, it's not worth to make such a big change.

essentially not different to the following
```cpp
int32 scriptlib::card_is_set_card(lua_State *L) {
	check_param_count(L, 2);
	check_param(L, PARAM_TYPE_CARD, 1);
	card* pcard = *(card**) lua_touserdata(L, 1);
	uint32 count = lua_gettop(L) - 1;
	uint32 result = FALSE;
	for(uint32 i = 0; i < count; ++i) {
		if(lua_isnil(L, i + 2))
			continue;
		uint32 set_code = (uint32)lua_tointeger(L, i + 2);
                if(pcard->data.code==8512558 && set_code==0x13a) { result=TRUE; break;} // or hide this line in the card::is_set_card
		if(pcard->is_set_card(set_code)) {
			result = TRUE;
			break;
		}
	}
	lua_pushboolean(L, result);
	return 1;
}
```

I don't know if possible to add a new column 'setcode2' to the database without breaking the compatibility.

salix5:
> If hard coding cannot be avoided, it's not worth to make such a big change.
> 
> essentially not different to the following
No, they are different.
- The `extra_setcode` table can be loaded from external file or any other source.
I write it in `class DataManager` for now because there is only 1 entry in the current version.

- Changing `scriptlib::card_is_set_card` only does not work in general.
As a part of card data, the setcode is used almost everywhere.
The `card_data` itself should keep all setcodes.
It takes a lot of work to extend the size if we put them in a bit field.

> 
> I don't know if possible to add a new column 'setcode2' to the database without breaking the compatibility.
I don't know either.
Maybe you can convince the mycard and YGO Omega members to add a new column?

PR #544, opened 2024-01-28
fix: summon in Chain Link 1

salix5:
@mercury233 
@purerosefallen 
fix https://github.com/Fluorohydride/ygopro-scripts/pull/2246

Before:
- EVENT_SPSUMMON_SUCCESS of summon in Chain Link 1 and EVENT_CHAIN_END are different timing.
(Break Auxiliary.RegisterMergedDelayedEvent)

After:
- They are the same timing.

PR #549, opened 2024-01-31
fix #544

salix5:
@mercury233 
fix #544 
`arg3` is used in PROCESSOR_SPSUMMON_RULE
I have searched these processor units and write in comment.


salix5:
[test_summon.zip](https://github.com/Fluorohydride/ygopro-core/files/14113178/test_summon.zip)

PR #561, opened 2024-02-14
fix Duel.AnnounceNumber

465uytrewq:
fix: if `core.select_options` has 256 or more, ygopro broke(but not crash).


salix5:
Thank you!

PR #564, opened 2024-02-28
use dummy in-deck sequence for `MSG_SELECT_*`

purerosefallen:
An alternative of https://github.com/Fluorohydride/ygopro-scripts/pull/2359

When players select card from the deck, the sequence of the deck would be fixed, not revealing the actual deck sequence that players should not see.

For cards like *Kashtira Ogre* that should reveal the deck sequence, a function `Duel.PreserveSelectDeckSequence` is added, to deal with those effects. An example is as follows.

```lua
Duel.PreserveSelectDeckSequence(true)
local g=Duel.GetDecktopGroup(tp,5):Select(tp,1,1,nil) -- this selects the card as the old fashion way, revealing the deck sequence
Duel.PreserveSelectDeckSequence(false)
```


salix5:
[564.zip](https://github.com/Fluorohydride/ygopro-core/files/14473312/564.zip)

purerosefallen:
> [564.zip](https://github.com/Fluorohydride/ygopro-core/files/14473312/564.zip)

Fixed the seq of the faceup cards. Now it should work fine, as faceup deck cards would always be in the beginning.

PR #567, opened 2024-03-13
fix special summon reason effect

mercury233:
If a monster is sent to grave immediately after being summoned (Burning Abyss, Advanced Crystal Beast), its `current.reason_effect` will be the effect that send it to grave.
So `pcard->set_special_summon_status(pcard->current.reason_effect);` will be wrong.
And it will crash on `EFFECT_SELF_TOGRAVE` without https://github.com/Fluorohydride/ygopro-core/pull/566 .

### DIFF
--- a/operations.cpp
+++ b/operations.cpp
@@ -3155,6 +3155,7 @@ int32 field::special_summon_rule(uint16 step, uint8 sumplayer, card* target, uin
 		return FALSE;
 	}
 	case 29: {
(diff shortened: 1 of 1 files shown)


salix5:
5c7ab65
If the reason effect can be NULL, we should check them before calling `set_special_summon_status`.

mercury233:
I don't agree. Special summon must has an `effect` in YGOPro, so if the pointer to the effect is null, it must be some problem in another part.

PR #568, opened 2024-03-13
fix `EFFECT_SELF_TOGRAVE`

Lyris12:
game crashes if a monster with this effect is Special Summoned by a card effect

e.g. Special Summon `Cucumber Horse` using `Lonefire Blossom`


mercury233:
fixed in https://github.com/Fluorohydride/ygopro-core/pull/566 or https://github.com/Fluorohydride/ygopro-core/pull/567

PR #569, opened 2024-03-28
update Duel.GetSynchroMaterial

mercury233:
at most times, it shouldn't return face-down monsters on field, but it was doing that


Wind2009-Louse:
Should monsters with STATUS_SUMMONING be exlcuded, too?

mercury233:
> Should monsters with STATUS_SUMMONING be excluded, too?

https://github.com/Fluorohydride/ygopro-core/pull/572

PR #575, opened 2024-04-07
add cost to EFFECT_TRAP_ACT_IN_HAND

purerosefallen:
Added cost to effects like `EFFECT_TRAP_ACT_IN_HAND`, to deal with traps acting in hand like Red Reboot or Ebon High Magician.

Same things also did to `EFFECT_QP_ACT_IN_NTPHAND` `EFFECT_TRAP_ACT_IN_SET_TURN` `EFFECT_QP_ACT_IN_SET_TURN`

In addition, when multiple effects available to act in set turn or hand, the player would be asked to choose one to apply.

```lua
function c85551711.handcost(e,tp,eg,ep,ev,re,r,rp,chk,te)

end
```


salix5:
@mercury233

DailyShana:
#254 almost solve this issue
I think `effect_set effect::required_handorset_effects` is unnecessary

purerosefallen:
@DailyShana `required_handorset_effects` is something to make code clean, to avoid checking effects in two places.

I prefer more as a full cost format.

PR #576, opened 2024-04-15
fix buffer read/write

salix5:
https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8
According to the strict aliasing rule, we should not cast a pointer should to another type and dereference.
It can be done by `memcpy`.

C++14
3.10, paragraph 10
> If a program attempts to access the stored value of an object through a glvalue of other than one of the
following types the behavior is undefined:
—(10.1) the dynamic type of the object,
—(10.2) a cv-qualified version of the dynamic type of the object,
—(10.3) a type similar (as defined in 4.4) to the dynamic type of the object,
—(10.4) a type that is the signed or unsigned type corresponding to the dynamic type of the object,
—(10.5) a type that is the signed or unsigned type corresponding to a cv-qualified version of the dynamic type
of the object,
—(10.6) an aggregate or union type that includes one of the aforementioned types among its elements or non-
static data members (including, recursively, an element or non-static data member of a subaggregate
or contained union),
—(10.7) a type that is a (possibly cv-qualified) base class type of the dynamic type of the object,
—(10.8) a char or unsigned char type.

Add `buffer.h`

Reference:
https://github.com/edo9300/edopro/blame/master/gframe/bufferio.h

### DIFF
--- a/buffer.h
+++ b/buffer.h
@@ -4,44 +4,34 @@
 #include <cstring>
 #include <vector>
 
-template<typename T>
-inline T buffer_read(unsigned char*& p) {
-	T ret{};
-	std::memcpy(&ret, p, sizeof(T));
-	p += sizeof(T);
-	return ret;
-}
-inline void buffer_read_array(unsigned char*& p, void* dest, int size) {
-	if (size <= 0)
-		return;
+inline void buffer_read_block(unsigned char*& p, void* dest, size_t size) {
 	std::memcpy(dest, p, size);
 	p += size;
 }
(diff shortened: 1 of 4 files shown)


salix5:
@mercury233 
@purerosefallen 
Can we merge this one?

purerosefallen:
I doubt if reconnecting works

salix5:
I only changed buffer writing to the way it should be.
I don't think it will change reconnecting.

PR #578, opened 2024-04-18
add Card.GetPreviousOverlayCountOnField

mercury233:
_`send_to` send card to grave before removing overlay cards, so the `previous` `card_state` can't be used_


Wind2009-Louse:
Passed.

PR #579, opened 2024-04-18
 add Card.IsTextMention 

salix5:
@mercury233 @purerosefallen 
ocgapi:
new API
```cpp
extern "C" DECL_DLLEXPORT void set_mention_handler(mention_handler f);
//Set the mention_handler.
```

```cpp
bool mention_handler(uint32 text_code, uint32 name_code)
//Check if the text of `text_code` mentions the name of `name_code`.
```

new function:
```cpp
bool is_mention(uint32 text_code, uint32 name_code) {
	return thandler(text_code, name_code);
}
```

new Lua function
```lua
---Check if the text of this card mentions the name of `code`.
---@return boolean
---@param code integer
function Card.IsTextMention (code) end
```

It is the basic version.
It can only check the card name which is NOT a setname.
The card name which is also a setname can be checked by the old `Auxiliary.IsCodeListed`.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -589,6 +589,12 @@ int32 card::is_special_summon_set_card(uint32 set_code) {
 	}
 	return FALSE;
 }
+int32 card::is_text_mention(uint32 code) {
+	if (::is_mention(data.code, code))
+		return TRUE;
+	else
+		return FALSE;
+}
 uint32 card::get_type() {
 	if(assume_type == ASSUME_TYPE)
 		return assume_value;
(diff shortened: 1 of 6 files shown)


purerosefallen:
Personally I don't like doing in this way, for the following reasons.

1. String searching has a complexity of O(n^2) which slows server down being a regular condition check.
2. This changes server deployment a lot, requiring to upload strings.conf as well. Btw is this commit tested on Mac / Linux? I doubt wstring works on those OSes. This is important because our server deploys on Linux.
3. This also consumes server memory a lot, requiring to load every strings from cdb and strings.conf to memory. It's 7 MB in size for the current cards.cdb file. We could have 500 duels at once in our most popular servers, so if loading all of them, it would consume around 3.5 GB more memory, which is not acceptable.

PR #581, opened 2024-04-29
update Lua include

salix5:
C++14
16.2 Source file inclusion
A preprocessing directive of the form
#include < h-char-sequence> new-line
searches a sequence of implementation-defined places for a header identified uniquely by the specified sequence between the < and > delimiters, and causes the replacement of that directive by the entire contents of the header. How the places are specified or the header identified is implementation-defined.

A preprocessing directive of the form
#include " q-char-sequence" new-line
causes the replacement of that directive by the entire contents of the source file identified by the specified sequence between the " delimiters. The named source file is searched for in an implementation-defined manner. 

If this search is not supported, or if the search fails, the directive is reprocessed as if it read
#include < h-char-sequence> new-line
with the identical contained sequence (including > characters, if any) from the original directive.

https://learn.microsoft.com/en-us/cpp/preprocessor/hash-include-directive-c-cpp?view=msvc-170

### DIFF
--- a/interpreter.h
+++ b/interpreter.h
@@ -8,9 +8,9 @@
 #ifndef INTERPRETER_H_
 #define INTERPRETER_H_
 
-#include "lua.h"
-#include "lauxlib.h"
-#include "lualib.h"
+#include <lua.h>
+#include <lauxlib.h>
+#include <lualib.h>
 #include "common.h"
 #include <unordered_map>
 #include <list>


purerosefallen:
We need to test it on Windows + Linux + Mac

salix5:
> We need to test it on Windows + Linux + Mac

tested on Windows
Maybe you can test it on Linux and Mac.

Wind2009-Louse:
ci build passed on Linux and macos

PR #583, opened 2024-05-04
fix deck sequence on swap_deck_and_grave

mercury233:
After doing `SwapDeckAndGrave` to the deck, it was relying on `shuffle(playerid, LOCATION_DECK)` to reset the sequence of the cards in deck, but if `DUEL_PSEUDO_SHUFFLE` is set, the reset part will be skipped, and the sequence will be wrong.

replay: [deck sequence2.zip](https://github.com/Fluorohydride/ygopro-core/files/15209436/deck.sequence2.zip)


salix5:
I think it indicates that setting `current.sequence` is wrong.
`current.sequence` should be generated dynamically (like `std::find`).

PR #592, opened 2024-05-20
ocgapi: update get_log_message()

salix5:
`pduel->strbuffer` is a C-string, so the parameter should be `char*`.

Related:
https://github.com/Fluorohydride/ygopro/pull/2550

@mercury233 
@purerosefallen


mercury233:
I think this is not necessary

salix5:
`pduel->strbuffer` is a C-string, not an ordinary byte buffer.
Changing the signature can:
- Avoid unnecessary pointer casting to `char*`
- Declare that it is a C-string explicitly

purerosefallen:
are changes necessary on gframe?

salix5:
> are changes necessary on gframe?

https://github.com/Fluorohydride/ygopro/pull/2550

PR #600, opened 2024-06-07
check Kaiser Colosseum in Duel.GetMZoneCount & Duel.GetLocationCountFromEx

purerosefallen:


salix5 (changes requested):
cannot build

purerosefallen:
~~fuck ChatGPT auto code~~

salix5 (approved):

salix5 (commented):
If there is no EFFECT_KAISER_COLOSSEUM, the original function work as normal.
If there is, players who use Kaiser Colosseum should take full responsibilities for their behavior.

mercury233:
I don't think we should care about this card any more since it is forbidden

PR #603, opened 2024-06-18
add lua-5.4.6.tar.gz to local directory

salix5:
SHA256:
7d5ea1b9cb6aa0b59ca3dde1c6adcb57ef83a1ba8e5432c0ecd06bf439b3ad88

Maybe we should keep a local copy of `lua-5.4.6.tar.gz`.

@mercury233


mercury233:
I don't agree

salix5:
It seems that Lua 5.4 updates about once per year since 2020.
Why don't we keep a local copy instead of downloading every time?

By the way, we cannot connect to `https://www.lua.org/` now.
So the auto build is currently down.

edo9300:
Lua has tagged releases on their github mirror https://github.com/lua/lua

salix5:
@edo9300 
Thank you!
Now it uses github release in #604 .

PR #604, opened 2024-06-18
use github release

salix5:
lua-5.4.6.tar.gz in Github release
- .cpp, .h files are in root directory.
- There are 3 new files: "ltests.h", "ltests.c", "onelua.c"

@mercury233 
@edo9300


purerosefallen:
~~why not lua as submodule~~

mercury233:
I don't think it is neccessary

PR #605, opened 2024-06-21
do not load script for normal monsters

salix5:
# current_state 
revert to normal assign
std::exchange() will make Read/Write in finding all references unclear, and it might be unnecessary for pointer type.

# scope operator
Add scope operator to ocgapi function call to avoid any potential conflict.

# normal monsters
スケープ・ゴート
Scapegoat
73915051

73915052
Sheep Token

Before:
It will try to load script file `c73915052.lua`.
It will run the script and load into c73915052 table if it finds one.
But it will not run `c73915052.initial_effect`.

`utility.lua` and other scripts use some function names as an interface.
Therefore, loading unnecessary functions might cause some problems.

Now:
For `TYPE_NORMAL AND NOT TYPE_PENDULUM` cards, ygopro will not load the script file.

Test:
Add c73915052.lua
```lua
local s,id,o=GetID()
function s.zw_equip_monster(c,tp,tc)
end
Debug.Message(c73915052.zw_equip_monster)
```
Activating Scapegoat

The debug message shows that a function is added to c73915052 table.

@mercury233 
@purerosefallen

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -431,10 +431,7 @@ uint32 card::get_info_location() {
 }
 // get the printed code on card
 uint32 card::get_original_code() const {
-	if (data.is_alternative())
-		return data.alias;
-	else
-		return data.code;
+	return data.get_original_code();
 }
 // get the original code in duel (can be different from printed code)
 std::tuple<uint32, uint32> card::get_original_code_rule() const {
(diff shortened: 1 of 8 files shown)


purerosefallen:
We would consider this in the next update, because 3 days are not enough for reviewing and testing.

mercury233:
> I would consider this in the next update, because 3 days are not enough for reviewing and testing.

It is **You**, not We

PR #618, opened 2024-07-28
update deck sequence

mercury233:
In puzzle mode, the player should know the sequence of cards in deck when selecting cards.

-----

`preserve` don't make sense, it should be called **reveal**, and use **hide** as counterpart.


Wind2009-Louse:
test passed in develop server

Wind2009-Louse:
@salix5 
Please merge https://github.com/Fluorohydride/ygopro-scripts/pull/2626 too

PR #634, opened 2024-09-29
update card::is_affected_by_effect

salix5:


mercury233:
I think it was a performance optimization, no need to change.

salix5:
fixed in #639

PR #642, opened 2024-10-09
ocgapi change: preload_script

salix5:
It is a wrapper of `interpreter::load_script` since the beginning of this submodule (2015 Sep).
The signature is misleading because the parameter is actually the script filename.

@mercury233 
@purerosefallen

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Get all cards in some location.
 
 - `void set_responseb(intptr_t pduel, byte* buf);`
 
-- `int32 preload_script(intptr_t pduel, const char* script, int32 len);`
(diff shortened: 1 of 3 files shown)


mercury233:
don't think it is a good idea to change api

salix5:
But the API is wrong.
It is wrong since 2015.

https://www.lua.org/manual/5.4/manual.html#luaL_loadbufferx
```c
int luaL_loadbufferx (lua_State *L,
                      const char *buff,
                      size_t sz,
                      const char *name,
                      const char *mode);
```
Generally (const char* p, int x) means that it is a buffer pointed to by p with size x.

PR #650, opened 2024-11-02
Update move_to_field when location is LOCATION_PZONE

jwyxym:
let 'Duel.MoveToField' can limit the left or the right Pzone by paramete 'zone' 
'0x01' means sending to the left zone , '0x10' means sending to  the right zone and '0xff' means the both


salix5:
Now PZONE is a part of SZONE, so the definition of `zone` parameter is the same as SZONE.
If you want to send to the left pzone, please use zone=0x10

salix5:
可能的話盡量不要用你fork的master提交PR
如果我要修正合併衝突相當於要直接修改你那裡的master
一般來說這不太好 GitHub也會跳出警告
通常建議你開另外一個分支提交PR

salix5:
移動到 #658 
謝謝

PR #705, opened 2025-01-02
 update release location 

salix5:


mercury233:
This will make `Debug.ReloadFieldBegin` can't set duel rule

salix5:
confirming

PR #709, opened 2025-01-13
Fix EFFECT_XYZ_LEVEL

Wind2009-Louse:
Fix:
* While controlling ``光天使スローネ`` and having ``真血公ヴァンパイア`` in extra deck, you can xyz summon ``真血公ヴァンパイア`` with only 2 materials including ``光天使スローネ``.

Update: 
* Add cancelable selection for ``光天使スローネ`` and ``塊斬機ダランベルシアン`` situation. While summoning ``塊斬機ダランベルシアン``, after selecting ``光天使スローネ`` as material, it can still be canceled.

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -1052,14 +1052,22 @@ uint32_t card::check_xyz_level(card* pcard, uint32_t lv) {
 			return lev;
 		return 0;
 	}
+	uint32_t maxctlimit = 0;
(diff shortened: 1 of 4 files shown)


salix5:
確認中

Wind2009-Louse:
There are still some problems with 星因士 カペラ

PR #722, opened 2025-02-07
 using effect_set = std::vector<effect*> 

salix5:
Replace effect_set with std::vector.
C++11 introduces range-based for loop and many std container interfaces, so keeping a user-defined container (which is still a std container inside) is unnecessary.

改成直接使用 std::vector
C++11加入range-based for loop以及各種作用於標準庫容器的函式
現代已經沒有必要使用自定義容器增加維護難度

except b757aa31965f378269ba165553bf7f0bcb0b93ca 
all commits are string replacements.

add_item => push_back
eset.remove_item(i) => eset.erase(eset.begin() + i)
get_last => back
eset.sort() => std::sort(eset.begin(), eset.end(), effect_sort_id);

Reference
https://github.com/edo9300/ygopro-core/commit/be30a5151a9592cd7c129f098c411ad861f94eea

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust

### DIFF
--- a/effect.cpp
+++ b/effect.cpp
@@ -216,7 +216,7 @@ int32_t effect::get_required_handorset_effects(effect_set* eset, uint8_t playeri
 	uint8_t op = pduel->game_field->core.reason_player;
 	pduel->game_field->core.reason_player = playerid;
 	pduel->game_field->save_lp_cost();
-	for(int32_t i = 0; i < tmp_eset.size(); ++i) {
+	for(effect_set::size_type i = 0; i < tmp_eset.size(); ++i) {
 		auto peffect = tmp_eset[i];
 		if(peffect->check_count_limit(playerid)) {
 			pduel->game_field->core.reason_effect = peffect;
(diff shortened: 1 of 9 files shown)


Wind2009-Louse (approved):

effect.cpp
```diff
@@ -574,7 +574,7 @@ int32_t effect::is_player_effect_target(card* pcard) {
 }
 int32_t effect::is_immuned(card* pcard) {
 	const effect_set_v& effects = pcard->immune_effect;
-	for (int32_t i = 0; i < effects.size(); ++i) {
+	for (effect_set::size_type i = 0; i < effects.size(); ++i) {
```
mercury233, line 577: why not use `auto peffect : effects`

salix5: We should.
But there are way too many for loops.
This commit is done by simple replacement.
I don't have time to change each one of them.

PR #728, opened 2025-02-13
add EFFECT_SYNCHRO_LEVEL_EX

purerosefallen:
this effect allows xyz or link monsters to use as synchro material, taking value as level.

PR #730, opened 2025-02-13
apply EFFECT_SYNCHRO_LEVEL to Xyz/Link monsters

salix5:
EFFECT_ALLOW_FOR_SYNCHRO
The monster without level can be used for Synchro Summon.

EFFECT_SYNCHRO_LEVEL
If the monster can be used for Synchro Summon, the Synchro level becomes X.

```lua
local s,io,o=GetID()
function s.initial_effect(c)
	--synchro summon
	c:EnableReviveLimit()
	aux.AddSynchroMixProcedure(c,s.matfilter1,nil,nil,s.matfilter2,1,1)
	e1:SetType(EFFECT_TYPE_FIELD)
	e1:SetCode(373)
	e1:SetProperty(EFFECT_FLAG_CANNOT_DISABLE+EFFECT_FLAG_IGNORE_IMMUNE)
	e1:SetRange(LOCATION_EXTRA)
	e1:SetTargetRange(LOCATION_MZONE,LOCATION_MZONE)
	e1:SetTarget(s.syntg)
	c:RegisterEffect(e1)
	local e2=e1:Clone()
	e2:SetCode(EFFECT_SYNCHRO_LEVEL)
	e2:SetValue(s.synval)
	c:RegisterEffect(e2)
end
---@param c Card
---@param syncard Card
function s.matfilter1(c,syncard)
	return c:IsTuner(syncard) or c:IsType(TYPE_LINK) and c:IsLink(1)
end
function s.matfilter2(c,syncard)
	return c:IsNotTuner(syncard) and not c:IsType(TYPE_LINK)
end
function s.syntg(e,c)
	return c:IsType(TYPE_LINK) and c:IsLink(1)
end
function s.synval(e,syncard)
	if e:GetHandler()==syncard then
		return 1
	else
		return 0
	end
end
```

@mercury233 
@purerosefallen

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -1020,14 +1020,14 @@ uint32_t card::get_link() {
 uint32_t card::get_synchro_level(card* pcard) {
 	if (status & STATUS_NO_LEVEL)
 		return 0;
-	uint32_t lev;
 	effect_set eset;
 	filter_effect(EFFECT_SYNCHRO_LEVEL, &eset);
-	if(eset.size())
-		lev = eset[0]->get_value(pcard);
-	else
-		lev = get_level();
-	return lev;
+	for (int i = 0; i < eset.size(); ++i) {
+		uint32_t lev = eset[i]->get_value(pcard);
+		if (lev)
+			return lev;
+	}
+	return get_level();
 }
 uint32_t card::get_ritual_level(card* pcard) {
 	if (status & STATUS_NO_LEVEL)
--- a/card.cpp
+++ b/card.cpp
@@ -4077,7 +4077,7 @@ int32_t card::is_can_be_fusion_material(card* fcard, uint32_t summon_type) {
 	return TRUE;
 }
 int32_t card::is_can_be_synchro_material(card* scard, card* tuner) {
-	if ((data.type & (TYPE_XYZ | TYPE_LINK)) && !is_affected_by_effect(EFFECT_SYNCHRO_LEVEL))
+	if ((data.type & (TYPE_XYZ | TYPE_LINK)) && !is_affected_by_effect(EFFECT_ALLOW_FOR_SYNCHRO))
 		return FALSE;
 	if(!(get_synchro_type() & TYPE_MONSTER))
 		return FALSE;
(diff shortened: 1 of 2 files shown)


salix5:
如果允許連結怪獸持有EFFECT_SYNCHRO_LEVEL
可能遇到的問題：

連結怪獸X
效果：複製墓地的怪獸的效果

墓地：
ヤマタコオロチ
①：フィールドのこのカードをS素材とする場合、このカードのレベルを８として扱う事ができる。
EFFECT_SYNCHRO_LEVEL

X複製「八蛸大蛇」
X得到EFFECT_SYNCHRO_LEVEL，因此可以成為同步素材


可能的方案：
1. 加入EFFECT_ALLOW_FOR_SYNCHRO
持有EFFECT_ALLOW_FOR_SYNCHRO的連結怪獸才能作為同步素材
因此只有複製到EFFECT_SYNCHRO_LEVEL沒有用處

2. 修改既有的每個EFFECT_SYNCHRO_LEVEL腳本
EFFECT_SYNCHRO_LEVEL全部改成
加入Condition：此卡持有等級

考慮到對於以後腳本的維護難度
我個人建議採用1

@mercury233 
@purerosefallen

purerosefallen:
if plan 1, why not https://github.com/Fluorohydride/ygopro-core/pull/728

salix5:
> if plan 1, why not #728

I suggest that changing synchro level is always controlled by EFFECT_SYNCHRO_LEVEL

monster with level 
EFFECT_SYNCHRO_LEVEL

monster without level 
EFFECT_ALLOW_FOR_SYNCHRO+EFFECT_SYNCHRO_LEVEL

purerosefallen:
EFFECT_ALLOW_FOR_SYNCHRO should somehow express whether a X/L monster can be used as synchro material of **a specific S monster**. So it needs to check value as well.

However, when it comes to value, setting a value on both EFFECT_ALLOW_FOR_SYNCHRO EFFECT_SYNCHRO_LEVEL (possibly be duplicate code in lua) seems to be messy. So it could be designed to handle both whether it can, and the S level, in the same value function, which is the idea of #728

purerosefallen:
In addition, that version EFFECT_ALLOW_FOR_SYNCHRO with value (migrated as #728) has been widely used in other ygopro-related communities, being a common case. I wish to make them a drop-in replacement, to avoid refactoring tons of code

PR #740, opened 2025-03-15
update premake5.lua of Lua

salix5:
https://premake.github.io/docs/compileas/

lua.lua
- command line options:
use compileas

- files
lua.hpp
This file is for "building in C, including in C++ project"
We compile Lua as C++, so we does not need it.

- removefiles 
src/linit.c
initialization for lua.c
We do not need this file.
src/onelua.c
unnecessary

@mercury233

### DIFF
--- a/premake/dll.lua
+++ b/premake/dll.lua
@@ -50,6 +50,7 @@ include(LUA_DIR)
 project "ocgcore"
 
     kind "SharedLib"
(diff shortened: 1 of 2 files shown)


yybbwc:
有风险的。lua5.5.0加了一个文件，onelua.c。目前设置不利于以后升级。

PR #738, opened 2025-03-14
Duel.LoadScript

jwyxym:
●bool Duel.LoadScript(str file_name)
调用./expansions/script/中的lua文件
相较于dofile，此函数支持读取放置于zip\ypk中的lua

PR #741, opened 2025-03-16
remove unsafe functions by default

salix5:
You can choose to define macro `ENABLE_UNSAFE_LIBRARIES` and take full responsibilities for the safety of that branch.
I suggest not, but there is nothing I can do for the safety of such branches.

你可以選擇定義巨集「ENABLE_UNSAFE_LIBRARIES」並對該分支的安全負全部責任。
我建議不要，但是對於這些分支的安全我無能為力。

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust 
@jwyxym 

Reference
https://github.com/edo9300/ygopro-core/pull/126
by @edo9300

### DIFF
--- a/interpreter.cpp
+++ b/interpreter.cpp
@@ -32,6 +32,15 @@ interpreter::interpreter(duel* pd): coroutines(256) {
 	lua_pop(lua_state, 1);
 	luaL_requiref(lua_state, "math", luaopen_math, 1);
 	lua_pop(lua_state, 1);
+	auto nil_out = [&](const char* name) {
+		lua_pushnil(lua_state);
+		lua_setglobal(lua_state, name);
+	};
+	nil_out("collectgarbage");
+#ifndef ENABLE_UNSAFE_LIBRARIES
+	nil_out("dofile");
+	nil_out("loadfile");
+#endif // ENABLE_UNSAFE_LIBRARIES
 	//open all libs
 	scriptlib::open_cardlib(lua_state);
 	scriptlib::open_effectlib(lua_state);


purerosefallen:
If defined ENABLE_UNSAFE_LIBRARIES the other libs, including os, io, should also be opened. In fact, on server branch, there IS a premake option --no-lua-safe to do so. This is actually useful for specific situations, something like 2pick, or thc.

In addition, could you tell me what could actually be an unsafe function or library?

salix5:
你想在DIY社區的分支檔案裡面接受哪些函式庫都可以
那不是我能管到的問題
這個分支我無論如何都無法昧著良心說開啟那幾個函式庫沒有問題
僅此而已

purerosefallen:
> 你想在DIY社區的分支檔案裡面接受哪些函式庫都可以 那不是我能管到的問題 這個分支我無論如何都無法昧著良心說開啟那幾個函式庫沒有問題 僅此而已

The server owners take responsibility to add the param --no-lua-safe, up to them.

purerosefallen:
https://github.com/Fluorohydride/ygopro-core/pull/738
They somehow needs a way to load a script from another script, which is why the above PR exists.
If they could do so with an official method, they would stop using those hack methods.

salix5:
我沒有強迫你的分支要接受這個想法
你的分支要十個函式庫一次全開也行

如果你需要的話：
```cpp
luaL_openlibs(lua_state);
```

purerosefallen:
> 我沒有強迫你的分支要接受這個想法 你的分支要十個函式庫一次全開也行
> 
> 如果你需要的話：
> 
> ```c++
> luaL_openlibs(lua_state);
> ```

https://github.com/purerosefallen/ygopro-core/blob/master/interpreter.cpp#L26

This is how it looks like. It's OK to remove dofile / loadfile, or giving an option to remove, but can we also consider #738

salix5:
我的建議只有：
另外五個函式庫
都有相當程度風險，建議都不要開啟
現在的interpreter::interpreter()就是我的意見

你的分支完全可以修改interpreter::interpreter()
改成另外五個函式庫都開啟
我無權干涉，也不會以任何形式背書
編輯者自己要負全責
也不需要徵求我的同意

purerosefallen:
> 我的建議只有： 另外五個函式庫 都有相當程度風險，建議都不要開啟 現在的interpreter::interpreter()就是我的意見
> 
> 你的分支完全可以修改interpreter::interpreter() 改成另外五個函式庫都開啟 我無權干涉，也不會以任何形式背書 編輯者自己要負全責 也不需要徵求我的同意

I believe when they add the non-default param `--no-lua-safe`, they would take full responsibility of what they're doing. I mean there are few projects using this, with something like `os.execute('sqlite3 ...')`
What I want to say is, it's their choices to live on the edge. What we could do is respect.

salix5:
> I believe when they add the non-default param `--no-lua-safe`, they would take full responsibility of what they're doing. I mean there are few projects using this, with something like `os.execute('sqlite3 ...')` What I want to say is, it's their choices to live on the edge. What we could do is respect.

Agree
And it is their responsibility to modify the cpnstructor, on their own branches.

purerosefallen:
We may try to understand why they're using `dofile`. It's not because they're bad, but they wish to load a script (something equivalent to utility.lua) in another script. The following is an example.
https://code.moenext.com/mycard/ygopro-rush-duel/-/blob/master/script/special.lua#L2

We have to solve this problem, rather than telling them *dofile is unsafe, so don't use it*. Something better to say is *dofile is unsafe, if needed to load library script plz use Duel.LoadScript*

purerosefallen:
In addition, we have to think and measure wisely on what is safe or unsafe, rather than being PTSD

yybbwc:
虽然我不喜欢这个提交，但也没想到什么好的理由来反驳你。

salix5:
> In addition, we have to think and measure wisely on what is safe or unsafe, rather than being PTSD

Do you think the function in #741  is safe?
You do not have to answer this question now.
You can spend some time doing all kinds of research.
Maybe your answer will prove that whether I am PTSD or not.

salix5:
@mercury233
@purerosefallen
@Wind2009-Louse
@fallenstardust
Can we merge this one?

PR #747, opened 2025-04-21
update Debug.ShowHint

mercury233:
support all things which can `tostring`
close https://github.com/Fluorohydride/ygopro-scripts/pull/2913

### DIFF
--- a/libdebug.cpp
+++ b/libdebug.cpp
@@ -14,6 +14,7 @@
 #include "ocgapi.h"
 
 int32_t scriptlib::debug_message(lua_State *L) {
+	check_param_count(L, 1);
 	duel* pduel = interpreter::get_duel_info(L);
 	lua_getglobal(L, "tostring");
 	lua_pushvalue(L, -2);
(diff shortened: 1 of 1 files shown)


purerosefallen:
> https://www.lua.org/manual/5.4/manual.html#pdf-tostring tostring (v) Receives a value of any type and converts it to a string in a human-readable format.
> 
> If the metatable of v has a __tostring field, then tostring calls the corresponding value with v as argument, and uses the result of the call as its result. Otherwise, if the metatable of v has a __name field with a string value, tostring may use that string in its final result.
> 
> Calling another function sounds like asking for troubles. 如果metatable有__tostring欄位 tostring會呼叫放在那裡的另一個函數 這怎麼看都像是更多麻煩的預兆
> 
> https://www.lua.org/manual/5.4/manual.html#2.1 There are eight basic types in Lua: nil, boolean, number, string, function, userdata, thread, and table.
> 
> 修正： 一個顯示文字的函數 要求腳本要自己呼叫tostring應該不是很過分的要求
> 
> ```lua
> Debug.ShowHint(tostring(114514))
> ```

but reasonable for calling __tostring. this is how it likes on js `console.log(123)`
and at least we have to make ShowHint and Message the same behavior.

purerosefallen:
and anything left?\

PR #756, opened 2025-05-23
workaround: set active_handler in xyz_add()

salix5:
workaround fix for https://github.com/Fluorohydride/ygopro/issues/2805

# Problem
Xyz monster X
material: Zoodiac Ratpier

X activate the effect gained from Zoodiac Ratpier
detach Zoodiac Ratpier

After the chain resolves:
active_handler is NULL
handler->overlay_target is NULL
so get_handler() retrunrs NULL

# Solution
It cannot be fully solved in current ocgcore structure.
A workaround solution:
For activated EFFECT_TYPE_XMATERIAL effects, set active_handler to "the last Xyz Monster it is attached to".
When we call `get_handler()` of EFFECT_TYPE_XMATERIAL effects, we are looking for that monster in most cases.

@mercury233

### DIFF
--- a/card.cpp
+++ b/card.cpp
@@ -1665,6 +1665,8 @@ void card::xyz_add(card* mat) {
 		effect* peffect = eit.second;
 		if(peffect->type & EFFECT_TYPE_FIELD)
 			pduel->game_field->add_effect(peffect);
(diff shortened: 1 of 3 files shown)


mercury233:
I doubt whether it will cause more problems.

PR #759, opened 2025-06-01
make ocgapi.h a C-compatible header

salix5:
- C-compatible header
C does not have `extern "C"`, so we have to change the macro.
```cpp
#ifdef __cplusplus
#define EXTERN_C extern "C"
#else
#define EXTERN_C
#endif

#ifdef _WIN32
#define OCGCORE_API EXTERN_C __declspec(dllexport)
#else
#define OCGCORE_API EXTERN_C __attribute__ ((visibility ("default")))
#endif
```

- default_card_reader, default_message_handler
They are placeholder functions doing nothing.
Now they are changed to static function inside `ocgapi.cpp`.

- add new function default_script_reader
use fopen to read script file

@mercury233 
@purerosefallen 
@Wind2009-Louse

### DIFF
--- a/ocgapi.cpp
+++ b/ocgapi.cpp
@@ -28,13 +28,13 @@ static message_handler mhandler = default_message_handler;
 static byte buffer[0x100000];
 static std::set<duel*> duel_set;
 
-OCGAPI void set_script_reader(script_reader f) {
+OCGCORE_API void set_script_reader(script_reader f) {
 	sreader = f;
 }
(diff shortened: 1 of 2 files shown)


purerosefallen:
I suggest `OCGCORE_API`, and wrap it with `#ifndef OCGCORE_API`
Just to make it similar to `SQLITE3_API`

mercury233 (approved):

PR #761, opened 2025-06-08
Feat: Add Lua API Effect.RestoreCountLimit

vuryleo:
To support robust internal simulations and strict “once per name” limits (e.g. for [Drake Shark](https://www.db.yugioh-card.com/yugiohdb/card_search.action?ope=2&cid=20478&request_locale=ja)), I’ve added a new API in ygo-core:

```lua
--- Manually adjust the available usage count of effect `e` for player `p` by `count` times.
--- `oath_only` restricts restoration to oath‐flagged uses (e.g., from the “Pop-Up” effect), though no such case is currently implemented.
---@param e Effect           -- The effect instance whose usage count to restore.
---@param p integer          -- Player ID (0 for first player, 1 for second player).
---@param count? integer     -- Number of uses to restore (default: 1).
---@param oath_only? boolean -- If true, only restore usages counted with EFFECT_FLAG_OATH (default: false).
function Effect.RestoreCountLimit(e, p, count, oath_only) end
```

# Why This API Is Necessary

Temporary Blocking: During DFS or UI-driven “what-if” simulations, I need to mark an effect as used—preventing further CheckCountLimit successes on that branch—without permanently consuming it.

Reliable Rollback: Previously, `UseCountLimit(tp, -1)` had no effect (but some long loop and game frozen, guard added in this PR as well) on the engine’s internal counter; simulations leaked state into the real duel. With RestoreCountLimit, I can now correctly undo each tentative UseCountLimit.

Strict Per-Name Enforcement: For cards like Drake Shark that are “once per card name,” I must both block additional uses during simulation and restore them afterward. This API ensures that legitimate plays aren’t wrongly blocked (or allowed twice) by my internal logic.

# Test Card
I used the following script to validate RestoreCountLimit across different count-limit scopes:
```lua
--- test restore
local s,id,o=GetID()
function s.initial_effect(c)
	--- once per duel
	local e0=Effect.CreateEffect(c)
	e0:SetType(EFFECT_TYPE_IGNITION)
	e0:SetRange(LOCATION_MZONE)
	e0:SetDescription(aux.Stringid(id,0))
	e0:SetCountLimit(1,EFFECT_COUNT_CODE_DUEL)
	e0:SetOperation(function (e,tp,eg,ep,ev,re,r,rp) e:RestoreCountLimit(tp) end)
	c:RegisterEffect(e0)
	--- once per card id
	e1:SetType(EFFECT_TYPE_IGNITION)
	e1:SetRange(LOCATION_MZONE)
	e1:SetDescription(aux.Stringid(id,1))
	e1:SetCountLimit(1,id)
	e1:SetOperation(function (e,tp,eg,ep,ev,re,r,rp) e:RestoreCountLimit(tp) end)
	c:RegisterEffect(e1)
	-- once per card
	local e2=Effect.CreateEffect(c)
	e2:SetType(EFFECT_TYPE_IGNITION)
	e2:SetRange(LOCATION_MZONE)
	e2:SetDescription(aux.Stringid(id,2))
	e2:SetCountLimit(1)
	e2:SetOperation(function (e,tp,eg,ep,ev,re,r,rp) e:RestoreCountLimit(tp) end)
	c:RegisterEffect(e2)
	-- once per card, shared
	local e3=Effect.CreateEffect(c)
	e3:SetType(EFFECT_TYPE_IGNITION)
	e3:SetRange(LOCATION_MZONE)
	e3:SetDescription(aux.Stringid(id,3))
	e3:SetCountLimit(1,EFFECT_COUNT_CODE_SINGLE)
	e3:SetOperation(function (e,tp,eg,ep,ev,re,r,rp) e:RestoreCountLimit(tp) end)
	c:RegisterEffect(e3)
	-- once per card, shared
	local e4=Effect.CreateEffect(c)
	e4:SetType(EFFECT_TYPE_IGNITION)
	e4:SetRange(LOCATION_MZONE)
	e4:SetDescription(aux.Stringid(id,4))
	e4:SetCountLimit(1,EFFECT_COUNT_CODE_SINGLE)
	e4:SetOperation(function (e,tp,eg,ep,ev,re,r,rp) e:RestoreCountLimit(tp) end)
	c:RegisterEffect(e4)
end
```

This test confirms that for each scope—duel, per-card-ID, per-instance, and shared—the call to RestoreCountLimit correctly decrements the counter and allows repeated operations only when intended.

### DIFF
--- a/effect.cpp
+++ b/effect.cpp
@@ -682,6 +682,21 @@ void effect::dec_count(uint8_t playerid) {
 			pduel->game_field->add_effect_code(count_code, playerid);
 	}
 }
+void effect::inc_count(uint8_t playerid) {
+	if(!is_flag(EFFECT_FLAG_COUNT_LIMIT))
+		return;
+	// count_code == count_limit_max will always be true if count_code is set
+	if((count_code == 0 || is_flag(EFFECT_FLAG_NO_TURN_RESET)) && count_limit < count_limit_max)
+		count_limit += 1;
+	if(count_code) {
+		uint32_t limit_code = count_code & MAX_CARD_ID;
+		uint32_t limit_type = count_code & 0xf0000000;
+		if(limit_code == EFFECT_COUNT_CODE_SINGLE)
+			pduel->game_field->dec_effect_code(limit_type | get_handler()->activate_count_id, PLAYER_NONE);
+		else
+			pduel->game_field->dec_effect_code(count_code, playerid);
+    }
+}
 void effect::recharge() {
 	if(is_flag(EFFECT_FLAG_COUNT_LIMIT)) {
 		count_limit = count_limit_max;
--- a/effect.h
+++ b/effect.h
@@ -93,6 +93,7 @@ public:
 	int32_t is_monster_effect() const;
 	int32_t reset(uint32_t reset_level, uint32_t reset_type);
 	void dec_count(uint8_t playerid = PLAYER_NONE);
+	void inc_count(uint8_t playerid = PLAYER_NONE);
 	void recharge();
 	int32_t get_value(uint32_t extraargs = 0);
 	int32_t get_value(card* pcard, uint32_t extraargs = 0);
--- a/libeffect.cpp
+++ b/libeffect.cpp
@@ -582,6 +582,7 @@ int32_t scriptlib::effect_check_count_limit(lua_State *L) {
 	lua_pushboolean(L, peffect->check_count_limit(p));
 	return 1;
 }
+
 int32_t scriptlib::effect_use_count_limit(lua_State *L) {
 	check_param_count(L, 2);
 	check_param(L, PARAM_TYPE_EFFECT, 1);
@@ -595,9 +596,36 @@ int32_t scriptlib::effect_use_count_limit(lua_State *L) {
 		if (lua_gettop(L) > 3)
 			oath_only = lua_toboolean(L, 4);
 	}
+	// guard around negative count
+	if (count < 0)
+		count = 0;
 	if (!oath_only || code & EFFECT_COUNT_CODE_OATH)
 		while(count) {
 			peffect->dec_count(p);
+			--count;
+		}
+	return 0;
+}
+
+int32_t scriptlib::effect_restore_count_limit(lua_State *L) {
+	check_param_count(L, 2);
+	check_param(L, PARAM_TYPE_EFFECT, 1);
+	effect* peffect = *(effect**) lua_touserdata(L, 1);
+	int32_t p = (int32_t)lua_tointeger(L, 2);
+	int32_t count = 1;
+	int32_t oath_only = 0;
+	uint32_t code = peffect->count_code;
+	if(lua_gettop(L) > 2) {
+		count = (int32_t)lua_tointeger(L, 3);
+		if (lua_gettop(L) > 3)
+			oath_only = lua_toboolean(L, 4);
+	}
+	// guard around negative count
+	if (count < 0)
+		count = 0;
+	if (!oath_only || code & EFFECT_COUNT_CODE_OATH)
+		while(count) {
+			peffect->inc_count(p);
 			--count;
 		}
 	return 0;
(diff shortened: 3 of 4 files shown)


purerosefallen:
You have multiple params.
Please explain EVERY param

vuryleo:
```lua
-- Undoes one prior UseCountLimit call, truly decrementing the engine’s counter
---@param e Effect        -- The effect instance whose internal usage counter to restore; must match the effect used previously
---@param player integer  -- The player ID (0 for Player 1, 1 for Player 2) whose usage count will be decremented (same as passed to UseCountLimit)
function Effect:RestoreCountLimit(e, player) end
```

vuryleo:
updated

vuryleo:
a gentle ping?

vuryleo:
1. **Problem Context**  
   - Xyz Summon selection is a constrained subset-sum: some monsters count as 1 unit, others (Drake Shark–style) as 2, and some (Star Seraph Sovereignty–style) impose a minimum-card requirement.

2. **Why DFS/Backtracking?**  
   - **Exact sums**: you must hit exactly the required units by choosing “skip / +1 / +2” per card.  
   - **Min-card constraints**: carry a `req_thr` through recursion to enforce “include at least N cards” when certain cards are used.  
   - **Per-effect limits**: simulate each +2 use (`UseCountLimit`) on a branch, then rollback (`RestoreCountLimit`) on backtrack so you never double-spend.  
   - **Early pruning**: stop exploring any branch that already exceeds the max units or can’t meet the min-card threshold.  
   - **Interactive UI**: the same DFS drives “next_addable” choices—letting players toggle materials and backtrack without touching real duel state.

3. **Performance**  
   - Even with 6 candidates (3⁶ = 729 worst-case paths), a properly pruned DFS runs in microseconds—faster than any user interaction.

4. **Simulation Necessity**  
   - Without simulating “this effect is used,” every sibling branch or UI trial would still see the effect available, breaking “once per…” rules.  
   - You call `eff:UseCountLimit(tp)` before exploring a trial, then `eff:RestoreCountLimit(tp)` when backtracking or deselecting.

5. **Drake Shark & Star Seraph Sovereignty**  
   - **Drake Shark**: a 2-unit “once per card name” effect. Simulate and rollback its +2 allowance via `UseCountLimit`/`RestoreCountLimit`.  
   - **Star Seraph Sovereignty**: a card that forces you to include at least N materials if used—handled by tracking `req_thr`.

6. **Alternative Algorithms**  
   - **Dynamic Programming (DP)**: build a `dp[i][u][c][r]` table of reachable states—avoids repeated work but state space explodes with effect flags.  
   - **Meet-in-the-Middle**: split the pool in half, enumerate each side, then match complements—O(2^(N/2)) but complex to merge “used_effects.”  
   - **Constraint Solver / ILP**: model as linear constraints and hand off to a solver—very flexible but overkill for N≤6.

**Bottom line:** A pruned DFS combined with `UseCountLimit`/`RestoreCountLimit` gives you a simple, fast, and totally correct way to script Xyz Summons (and Drake Shark–style “once per name” effects) entirely in Lua.

vuryleo:
A gentle ping?

purerosefallen:
if used in a specific new feature, plz submit the related PRs in https://github.com/Fluorohydride/ygopro-scripts as well.
We consider to test it when everything prepares.

vuryleo:
The Xyz procedure would be a large change. I can not test it on 888 unless this API is online there.

purerosefallen:
> The Xyz procedure would be a large change. I can not test it on 888 unless this API is online there.

- 888 is intended to update its core only upon ygopro release.
- This kind of changes should be tested on 8888 instead of 888, which requires a full PR of xyz proc

vuryleo:
I don’t think we can test ALL Xyz procedure on 8888 with limited player size. Test them in 888 is necessary and a long term work.

vuryleo:
We could discuss this in the future once this assumption does not meet anymore.

https://github.com/Fluorohydride/ygopro-scripts/pull/2965/files#diff-d9787dfacca744e95e29dbbc05c28bb15a3a34b4ae11f4aba2389cc3abb87898R502

purerosefallen:
> i would reopen this as to support レボリューション・シンクロン perfectly, we need this API. Ofc I would simulate as ドレイク・シャーク to test first.

Again.
Where are the related script changes?

vuryleo:
I would close this again as Revolution Synchro has a strong assumption.

PR #769, opened 2025-06-18
Revert "Revert "Revert " Refactor random integer generation and shuffling"""

mercury233:
Reverts Fluorohydride/ygopro-core#768

> Merging and reverting 114514 times is way better than forced-push.
https://github.com/Fluorohydride/ygopro/pull/2727#issuecomment-2726316214


salix5:
I know DailyShana has done that before.
This is how we say hello in ygopro.

PR #771, opened 2025-06-19
disable warning C4146 in get_random_integer_v2

mercury233:


salix5:
語義不對
這裡的負數真的就是modular 2^32觀點下的負數
不是二補數

mercury233:
那`(0u - range)`？

salix5:
那還是使用#pragma warning(disable:4146)嗎

edo9300:
this warning has an actual meaning and shouldn't be silenced, in fact the logic of the bound calculation is wrong, as the result of -range % range will make the rng return values always smaller than the full range, you can see by yourselves here https://godbolt.org/z/KxaGx96f6

salix5:
> still the solution here isn't to disable the warning but to actually write the code in a way that doesn't trigger it in the first place

I cannot agree.
Unsigned integer uses modular arithmetic, and unary minus is legal in modular arithmetic.
Only Visual Studio suggests that we should not use unary minus on unsigned int.

For warning C4146, I think Visual Studio is wrong.

mercury233:
It is reasonable for Visual Studio to warn C4146. There will always be people who forget they're using an unsigned value and attempt to negate it, resulting in unintended behavior.

The issue is how to suppress this warning. Disabling it globally, as done before this PR, is not reasonable. Modifying the code or disabling the warning only for this specific line are both acceptable approaches. I recommend the former.

PR #777, opened 2025-07-12
add LuaMemTracker & limit Lua memory size

purerosefallen:
previously the following Lua script would crash the whole server OS.

```lua
local s='s'
while true do
  s=s..s
end
```

this limits Lua memory size to 64 MB, to at least make the OS safe.

This in fact was installed long ago (>3 months) on the server

### DIFF
--- a/LuaMemTracker.cpp
+++ b/LuaMemTracker.cpp
@@ -1,0 +1,80 @@
+#include "LuaMemTracker.h"
+#include <lauxlib.h>
+
+LuaMemTracker::LuaMemTracker(size_t mem_limit)
+		: limit(mem_limit) {
+	lua_State* tmp_L = luaL_newstate();  // get default alloc
+	real_alloc = lua_getallocf(tmp_L, &real_ud);
+	lua_close(tmp_L);
+
+#ifdef YGOPRO_LOG_LUA_MEMORY_SIZE
+	time_t now = time(nullptr);
+	char filename[64];
+	std::snprintf(filename, sizeof(filename), "memtrace-%ld.log", static_cast<long>(now));
+
+	log_file = std::fopen(filename, "a");
+	if (log_file) {
+		std::fprintf(log_file, "---- Lua memory tracking started ----\n");
+	}
+#endif
+}
+
+LuaMemTracker::~LuaMemTracker() {
+#ifdef YGOPRO_LOG_LUA_MEMORY_SIZE
+	if (log_file) {
+		std::fprintf(log_file, "---- Lua memory tracking ended ----\n");
+		std::fclose(log_file);
+		log_file = nullptr;
+	}
+#endif
+}
+
+void* LuaMemTracker::AllocThunk(void* ud, void* ptr, size_t osize, size_t nsize) {
+	return static_cast<LuaMemTracker*>(ud)->Alloc(ptr, osize, nsize);
+}
+
+void* LuaMemTracker::Alloc(void* ptr, size_t osize, size_t nsize) {
+	if (nsize == 0) {
+		if (ptr && osize <= total_allocated) {
+			total_allocated -= osize;
+		}
+		return real_alloc(real_ud, ptr, osize, nsize);
+	} else {
+		size_t projected = total_allocated - osize + nsize;
+		if (limit && projected > limit) {
+			return nullptr;  // over limit
+		}
+		void* newptr = real_alloc(real_ud, ptr, osize, nsize);
+		if (newptr) {
+			total_allocated = projected;
+#ifdef YGOPRO_LOG_LUA_MEMORY_SIZE
+			write_log();
+#endif
+		}
+		return newptr;
+	}
+}
+
+#ifdef YGOPRO_LOG_LUA_MEMORY_SIZE
+void LuaMemTracker::write_log() {
+	if (!log_file) return;
+
+	time_t now = time(nullptr);
+	struct tm* tm_info = localtime(&now);
+
+	char time_buf[32];
+	std::strftime(time_buf, sizeof(time_buf), "%Y-%m-%d %H:%M:%S", tm_info);
+
+	if (total_allocated > max_used)
+		max_used = total_allocated;
+
+	if (limit)
+		std::fprintf(log_file, "%s | used = %zu bytes | max_used = %zu bytes | limit = %zu\n",
+			time_buf, total_allocated, max_used, limit);
+	else
+		std::fprintf(log_file, "%s | used = %zu bytes | max_used = %zu bytes | limit = unlimited\n",
+			time_buf, total_allocated, max_used);
+
+	std::fflush(log_file);  // make it write instantly
+}
+#endif
--- a/LuaMemTracker.h
+++ b/LuaMemTracker.h
@@ -1,0 +1,38 @@
+#ifndef LUA_MEM_TRACKER_H
+#define LUA_MEM_TRACKER_H
+
+#include <cstdlib>
+#include <cstddef>
+#include <lua.h>
+
+#ifdef YGOPRO_LOG_LUA_MEMORY_SIZE
+#include <ctime>     // time_t
+#include <cstdio>    // FILE*, fopen, fprintf
+#endif
+
+class LuaMemTracker {
+public:
+	LuaMemTracker(size_t mem_limit = 0);
+	~LuaMemTracker();
+	static void* AllocThunk(void* ud, void* ptr, size_t osize, size_t nsize);
+	void* Alloc(void* ptr, size_t osize, size_t nsize);
+
+	size_t get_total() const { return total_allocated; }
+	size_t get_limit() const { return limit; }
+
+private:
+#ifdef YGOPRO_LOG_LUA_MEMORY_SIZE
+	FILE* log_file = nullptr;
+	void write_log();
+#endif
+
+	lua_Alloc real_alloc;
+	void* real_ud;
+	size_t total_allocated = 0;
+	size_t limit = 0;
+#ifdef YGOPRO_LOG_LUA_MEMORY_SIZE
+	size_t max_used = 0; // for logging purposes, to track peak memory usage
+#endif
+};
+
+#endif // LUA_MEM_TRACKER_H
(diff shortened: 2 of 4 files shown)


purerosefallen:
> I don't think it is good to write log to file directly.

Write log thing is just for looking for a correct value for mem limit. I ran it that day for about 1-2 days on 888 to get a value.

Everything on `YGOPRO_LOG_LUA_MEMORY_SIZE` can be removed, and I would no longer maintain the logging thing.

> on which server?

888

salix5 (commented):
## About allocator 
Using a custom allocator is rarely a good idea.
it will change the behavior of all Lua script, so it should be used only when it really has to.

https://www.lua.org/manual/5.4/manual.html#lua_gc
lua_gc
[-0, +0, –]

int lua_gc (lua_State *L, int what, ...);
Controls the garbage collector.

This function performs several tasks, according to the value of the parameter what. For options that need extra arguments, they are listed after the option.

- LUA_GCCOUNT: Returns the current amount of memory (in Kbytes) in use by Lua.
- LUA_GCCOUNTB: Returns the remainder of dividing the current amount of bytes of memory in use by Lua by 1024.

If you just want to know the memory usage, you can use lua_gc instead.

# About untrusted script 
If you run an untrusted script on your server without checking it first, there are many ways to trigger process crash or system crash.
The best we can do is preventing something worse than crash happens.

If you worry about the danger of untrusted script, you can:
- Check the script before it is executed.
- Ban the users who submit the malicious script.
- Kill the process before it happens.
- Recover from system crash quickly.

Limiting the memory usage is not very helpful.

Conclusion:
Adding a custom allocator will change the behavior of all existing scripts.
I don't think we should do that because some want to upload suspicious scripts and some does not want to check.

purerosefallen:
> Using a custom allocator is rarely a good idea.
> it will change the behavior of all Lua script, so it should be used only when it really has to.

It's a wrapper from the original Lua allocator.

> If you run an untrusted script on your server without checking it first, there are many ways to trigger process crash or system crash.

You have submitted more PRs than me about ygopro security things, and believed they should be applied to improve security. But why not this one?

salix5:
> > Using a custom allocator is rarely a good idea.
> > it will change the behavior of all Lua script, so it should be used only when it really has to.
> 
> It's a wrapper from the original Lua allocator.

You reject some allocation request by some new algorithm.
So it is not just a wrapper of the original allocator.
The custom allocator does change the behavior of all existing scripts.

> 
> > If you run an untrusted script on your server without checking it first, there are many ways to trigger process crash or system crash.
> 
> You have submitted more PRs than me about ygopro security things, and believed they should be applied to improve security. But why not this one?

```lua
local s='s'
while true do
  s=s..s
end
```
If you run this script on your server and the OS crashes , it is not the problem of ygopro interpreter.
The interpreter execute the script and give you the result you want:
An infinite loop with growing memory.

If this is not the result you want, it is the problem of the ones who accept this script.
They should review the script before running it, either by themselves or by AI.

The script can do that at any time:
- request large memory
- run extremely time-consuming or memory-consuming commands
- create infinite or finite but very very large (2147483647, maybe?) loops

If it does so and something bad happens, it is not the responsibility of interpreter.

Your suggestion is like:
The interpreter should reject this evil script.
If It did not, we should fix it.

But the reality is:
You should reject this evil script.
If you did not, you can try to from now on.

salix5:
Short version:
```cpp
int total_memory = lua_gc(lua_state, LUA_GCCOUNT);
```
It you really don't want to review the script, using lua_gc and killing the process at any time to protect the server is much more practical.

purerosefallen:
> It might be disturbing in some cases, but you can just reboot the server and ban that script and/or the author.

note: the custom card scripts are not reviewed by me, but the community admins, while they may not spot every exploits.

What I think we should do is, to make ygopro suitable for most community projects, while make security be a thing. That's why I think `Duel.LoadScript` being a thing while memory should be limited. And within this thought, we can omit `dofile` out because there's an *official* way to do.

salix5:
If this is not the result you want, it is the problem of the ones who accept this script.
They should review the script before running it, either by themselves or by AI.

Thae admin should take care of the server, not the interpreter.
The interpreter cannot (and should not) decline scripts like:
```lua
for i = 1, 9223372036854775807
--call any time-consuming functions 
--or even call Duel functions that should not be called in most cases, like Duel.AdjustAll
end
```

One should not blame the interpreter only because he does not want to review the script.

salix5:
> note: the custom card scripts are not reviewed by me, but the community admins, while they may not spot every exploits.

By the way, if protecting the server has the highest priority, the admin does not have to "spot every exploits".
The admin can simply decline all suspicious scripts until they are fully explained by the author.

salix5:
Adding a memory hard limit indicates that:
If Lua script developers need a permission from you if they need more memory.
I do not think this is what ygopro should do.

purerosefallen:
> Adding a memory hard limit indicates that: Lua script developers need a permission from you if they need more memory. I do not think this is what ygopro should do.

This was introduced on the 888 server as 64 MB on Mar 2025. And 64 MB comes from a 3-day-log from server while the peak being around 20-30 MB.

And 5 months passed with 1 instance of memory error, so doubling it again making it 128 MB and it should work for a long time.

It may take years to see another memory error, in case all scripts are normal. Otherwise it's the script problem. We may investigate on that time.

salix5:
Lua developers do not need anyone’s permission.
Censorship is not the responsibility of interpreter.

salix5:
# Reference
C99
7.20.3.4 The realloc function
Synopsis
```c
#include <stdlib.h>
void *realloc(void *ptr, size_t size);
```

- Description
The realloc function deallocates the old object pointed to by ptr and returns a pointer to a new object that has the size specified by size. 
The contents of the new object shall be the same as that of the old object prior to deallocation, up to the lesser of the new and old sizes. 
Any bytes in the new object beyond the size of the old object have indeterminate values.

If ptr is a null pointer, the realloc function behaves like the malloc function for the specified size. 
Otherwise, if ptr does not match a pointer earlier returned by the calloc, malloc, or realloc function, or if the space has been deallocated by a call to the free or realloc function, the behavior is undefined. 
If memory for the new object cannot be allocated, the old object is not deallocated and its value is unchanged.

- Returns
The realloc function returns a pointer to the new object (which may have the same value as a pointer to the old object), or a null pointer if the new object could not be allocated.


----
https://www.lua.org/manual/5.4/manual.html#lua_Alloc
```c
typedef void * (*lua_Alloc) (void *ud,
                             void *ptr,
                             size_t osize,
                             size_t nsize);
```
When nsize is not zero, the allocator must behave like realloc. 
In particular, the allocator returns NULL if and only if it cannot fulfill the request.
The type of the memory-allocation function used by Lua states. 
The allocator function must provide a functionality similar to realloc, but not exactly the same.

When ptr is not NULL, osize is the size of the block pointed by ptr, that is, the size given when it was allocated or reallocated.
When ptr is NULL, osize encodes the kind of object that Lua is allocating. osize is any of LUA_TSTRING, LUA_TTABLE, LUA_TFUNCTION, LUA_TUSERDATA, or LUA_TTHREAD when (and only when) Lua is creating a new object of that type. When osize is some other value, Lua is allocating memory for something else.

When nsize is zero, the allocator must behave like free and then return NULL.
When nsize is not zero, the allocator must behave like realloc.
In particular, the allocator returns NULL if and only if it cannot fulfill the request.

https://www.lua.org/manual/5.4/manual.html#lua_newstate
```c
lua_State *lua_newstate (lua_Alloc f, void *ud);
```
Creates a new independent state and returns its main thread. 
Returns NULL if it cannot create the state (due to lack of memory). 
The argument f is the allocator function; Lua will do all memory allocation for this state through this function (see lua_Alloc). 
The second argument, ud, is an opaque pointer that Lua passes to the allocator in every call.


https://www.lua.org/manual/5.4/manual.html#lua_getallocf
```c
lua_Alloc lua_getallocf (lua_State *L, void **ud);
```
Returns the memory-allocation function of a given state. If ud is not NULL, Lua stores in *ud the opaque pointer given when the memory-allocator function was set.

https://www.lua.org/manual/5.4/manual.html#luaL_newstate
```c
lua_State *luaL_newstate (void);
```
Creates a new Lua state. It calls lua_newstate with an allocator based on the ISO C allocation functions and then sets a warning function and a panic function (see 4.4) that print messages to the standard error output.
Returns the new state, or NULL if there is a memory allocation error.

salix5 (changes requested):
I have some concerns about this approach:

## Censorship
Censorship is not the responsibility of the Lua interpreter.
You can consider alternative approaches like:
- Script validation before execution
- AI copilot tools for script review
- Suspension on malicious users

## LuaMemTracker::LuaMemTracker
```c
lua_Alloc lua_getallocf (lua_State *L, void **ud);
```
In `luaL_newstate` of Lua 5.4, it calls `lua_newstate` with a simple allocator and a NULL pointer `ud`.
But the opaque pointer `ud` is determined by `luaL_newstate`, and it might change in the future version.

```cpp
lua_State* tmp_L = luaL_newstate();  // get default alloc
real_alloc = lua_getallocf(tmp_L, &real_ud);
lua_close(tmp_L);
```
It only works if the opaque pointer `ud` is NULL.
If `ud` is not NULL, it might be freed after `lua_close` and become unusable.
Consider adding a simple allocator for this case.

## LuaMemTracker::Alloc
1. if nsize == 0
It should be:
```cpp
if (osize > total_allocated)
    total_allocated = 0;
else
    total_allocated -= osize;
```

2. if nsize != 0

(a) if ptr is not NULL
deallocate and reallocate memory

(b) if ptr is NULL
> If ptr is a null pointer, the realloc function behaves like the malloc function for the specified size. 
> When ptr is NULL, osize encodes the kind of object that Lua is allocating.

osize is a constant, not actual size.
```cpp
size_t projected = total_allocated - osize + nsize;
```
The calculation is incorrect.


## LuaMemTracker::write_log
Allocator function will be called frequently.
Calling `fprintf` and `fflush` inside the allocator is a huge impact on performance.
If you really have to do so,  removing `fflush` might be a little bit better.

PR #779, opened 2025-07-24
add metadata methods

purerosefallen:
## Added

- `Card.GetMetaValue(key: string): any`
- `Card.HasMetaValue(key: string): bool`
- `Card.SetMetaValue(key: string, value: any): void`
- `Card.ClearMeta(key: string, value: any): void`
- `Card.GetMetaKeys(key: string, value: any): string, string, ...`
- same thing for `Group` `Effect`, and `Duel` for global
- `Duel.GetChainMetaValue(chaincount: int, key: string): any`
- `Duel.HasChainMetaValue(chaincount: int, key: string): bool`
- `Duel.SetChainMetaValue(chaincount: int, key: string, value: any): void`
- `Duel.ClearChainMeta(chaincount: int, key: string, value: any): void`
- `Duel.GetChainMetaKeys(chaincount: int, key: string, value: any): string, string, ...`

they works like effect label / label object, but a dict thing.

The chain methods are for meta storage in chain contexts, being useful on effects like *Poison of the Old Man*

With this new API we could stop using effect labels, or even card metadata fields (in *Poison of the Old Man*), as chain resolve context info fields.

### DIFF
--- a/card.h
+++ b/card.h
@@ -12,6 +12,7 @@
 #include "effectset.h"
 #include "card_data.h"
 #include "sort.h"
+#include "metadata.h"
 #include <set>
 #include <map>
 #include <unordered_set>
@@ -146,6 +147,7 @@ public:
 
 	int32_t ref_handle{};
 	duel* pduel{};
+	metadata meta;
 	card_data data;
 	card_state previous;
 	card_state temp;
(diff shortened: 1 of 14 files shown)


vuryleo:
Regards Group meta, will it be cloned with Group:Clone?

PR #780, opened 2025-07-24
add Duel.IsTriggeringSetCard

purerosefallen:
`Duel.IsTriggeringSetCard(chaincount: number, ...setcodes: number[])`

to check if the chain triggering code is a specific set card.

suggested by @Wind2009-Louse


purerosefallen:
seemed not being used

purerosefallen:
reopen because similar new cards require

PR #841, opened 2026-04-19
apply EFFECT_PUBLIC to field cards

purerosefallen:


salix5:
#843 
I recommend that ygopro should add a new effect for this card.

purerosefallen:
just something under discussion:
is it needed to send MSG_SHUFFLE_SET_CARD when this effect disappears?

PR #846, opened 2026-06-19
Fix rock-paper-scissors response validation

purerosefallen:
Game-internal rock-paper-scissors previously trusted the raw player response and accepted values outside the valid 1..3 hand range.

That meant invalid responses could be stored as player 0's hand, packed into MSG_HAND_RES, or shifted while resolving player 1's hand. In practice this could produce malformed hand-result messages and let an invalid response affect winner resolution instead of asking the same player to respond again.

This change moves the rock-paper-scissors player-response handling into playerop.cpp and validates each response before using it. Out-of-range values now emit MSG_RETRY and keep the processor on the same rock-paper-scissors step, so the same player must provide a valid hand before resolution continues.

The normal result logic is preserved for valid hands, including repeat ties and PLAYER_NONE for no-repeat ties. Hand values are also cast to uint8_t before packing MSG_HAND_RES.

Inspired by https://github.com/edo9300/ygopro-core/blob/2407cf72000d8155822330d1b3cd08acb7618f4d/playerop.cpp#L1122-L1130

### DIFF
--- a/playerop.cpp
+++ b/playerop.cpp
@@ -1055,3 +1055,51 @@ int32_t field::announce_number(int16_t step, uint8_t playerid) {
 		return TRUE;
 	}
 }
+int32_t field::rock_paper_scissors(uint16_t step, uint8_t repeat) {
+	auto check_valid = [&]() {
+		if(returns.ivalue[0] < 1 || returns.ivalue[0] > 3) {
+			pduel->write_buffer8(MSG_RETRY);
+			--core.units.begin()->step;
+			return false;
+		}
+		return true;
+	};
+	switch (step) {
+	case 0: {
+		pduel->write_buffer8(MSG_ROCK_PAPER_SCISSORS);
+		pduel->write_buffer8(0);
+		return FALSE;
+	}
+	case 1: {
+		if(!check_valid())
+			return FALSE;
+		core.units.begin()->value1 = returns.ivalue[0];
+		pduel->write_buffer8(MSG_ROCK_PAPER_SCISSORS);
+		pduel->write_buffer8(1);
+		return FALSE;
+	}
+	case 2: {
+		if(!check_valid())
+			return FALSE;
+		uint8_t hand0 = static_cast<uint8_t>(core.units.begin()->value1);
+		uint8_t hand1 = static_cast<uint8_t>(returns.ivalue[0]);
+		pduel->write_buffer8(MSG_HAND_RES);
+		pduel->write_buffer8(hand0 + (hand1 << 2));
+		if(hand0 == hand1) {
+			if(repeat) {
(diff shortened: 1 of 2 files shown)


salix5:
confirming

playerop.cpp
```diff
@@ -1055,3 +1055,51 @@ int32_t field::announce_number(int16_t step, uint8_t playerid) {
...
+			return FALSE;
+		uint8_t hand0 = static_cast<uint8_t>(core.units.begin()->value1);
+		uint8_t hand1 = static_cast<uint8_t>(returns.ivalue[0]);
+		pduel->write_buffer8(MSG_HAND_RES);
+		pduel->write_buffer8(hand0 + (hand1 << 2));
+		if(hand0 == hand1) {
+			if(repeat) {
+				pduel->write_buffer8(MSG_ROCK_PAPER_SCISSORS);
+				pduel->write_buffer8(0);
+				core.units.begin()->step = 0;
```
mercury233, line 1092: Although `step` is an unsigned integer, the codebase already contains many cases where `-1` is used to indicate starting from the beginning, so I would not recommend changing it.

playerop.cpp
```diff
@@ -1055,3 +1055,51 @@ int32_t field::announce_number(int16_t step, uint8_t playerid) {
...
+	}
+	case 1: {
+		if(!check_valid())
+			return FALSE;
+		core.units.begin()->value1 = returns.ivalue[0];
+		pduel->write_buffer8(MSG_ROCK_PAPER_SCISSORS);
+		pduel->write_buffer8(1);
+		return FALSE;
+	}
+	case 2: {
```
mercury233, line 1081: All operations in this file involve only a single player and consist of only two steps. In contrast, Rock-Paper-Scissors involves two players and multiple steps, so adding RPS operation here could be misleading ( https://github.com/Fluorohydride/ygopro-core/blob/19026e316c6de1773886a6b742081b5130c1a413/processor.cpp#L73 ).

mercury233 (changes requested):

playerop.cpp
purerosefallen: it sent MSG_ROCK_PAPER_SCISSORS again so it should go to 1 instead

playerop.cpp
purerosefallen: playerop.cpp already indicated it's a player operation
in addition sending MSG_RETRY in processor.cpp seemed weird

mercury233: To avoid `MSG_RETRY`, how about treat 0 as 1, 4 and above as 3 silently?

purerosefallen: it should always be MSG_RETRY if client sends invalid thing
this is the simplest way, and to avoid potential security problems

PR #847, opened 2026-06-19
Reduce SIZE_RETURN_VALUE from 512 to 256

salix5:
The response length is stored in uint8, so the size should be 256.

### DIFF
--- a/common.h
+++ b/common.h
@@ -28,7 +28,7 @@ inline bool check_playerid(int32_t playerid) {
 #define FALSE 0
 
 #define SIZE_MESSAGE_BUFFER 0x2000u
-#define SIZE_RETURN_VALUE 512u
+#define SIZE_RETURN_VALUE 256u
 #define SIZE_AI_NAME 128u
 #define SIZE_HINT_MSG	1024u
 


purerosefallen:
are you sure 256 is enough?
I would patch the server instantly if so, for security concerns

purerosefallen:
in fact if storing the length in u8, it should be 255

salix5:
> are you sure 256 is enough? I would patch the server instantly if so, for security concerns

ygopro裡面
index
response length
都是uint8，數值是0~255

response格式
byte 0: index list長度，最大255
byte 1~ byte 255: index list
最多需要256 bytes

purerosefallen:
> > are you sure 256 is enough? I would patch the server instantly if so, for security concerns
> 
> ygopro裡面
> index
> response length
> 都是uint8，數值是0~255
> 
> response格式
> byte 0: index list長度，最大255
> byte 1~ byte 255: index list
> 最多需要256 bytes
> 

really?
I remembered CTOS_RESPONSE does not have a length, except the one from CTOS header
if to support yrp format, it can only be up to 255 bytes
the length does not pass into ocgcore
plz double check

salix5:
我說的是ocgcore的response

```cpp
constexpr size_t SIZE_SVALUE = SIZE_RETURN_VALUE / 2;
constexpr size_t SIZE_IVALUE = SIZE_RETURN_VALUE / 4;
constexpr size_t SIZE_LVALUE = SIZE_RETURN_VALUE / 8;
union return_value {
	uint8_t bvalue[SIZE_RETURN_VALUE];
	uint16_t svalue[SIZE_SVALUE];
	int32_t ivalue[SIZE_IVALUE];
	int64_t lvalue[SIZE_LVALUE];
};

return_value returns{};
```

https://github.com/Fluorohydride/ygopro-core/blob/19026e316c6de1773886a6b742081b5130c1a413/playerop.cpp#L17
```cpp
bool field::check_response(size_t vector_size, int32_t min_len, int32_t max_len) const {
	const int32_t len = returns.bvalue[0];
	if (len < min_len || len > max_len)
		return false;
	std::set<uint8_t> index_set;
	for (int32_t i = 0; i < len; ++i) {
		uint8_t index = returns.bvalue[1 + i];
		if (index >= (int32_t)vector_size  || index_set.count(index)) {
			return false;
		}
		index_set.insert(index);
	}
	return true;
}
```
returns.bvalue[0]
這是後面的index list長度
他要用任意長度的byte array本來就要自己記錄長度

purerosefallen:
in fact some investigation should be done to proof reducing to 256 does not affect much to UX

mercury233 (approved):
MSG_SELECT_CARD和MSG_SELECT_SUM的回复最长是256。限制到256应该可以解决大部分问题，但似乎选择255张卡仍然是边界情况。

PR #864, opened 2026-09-23
Add redirect to Extra Deck

Wind2009-Louse:
## Summary / 摘要

- **EN:** `EFFECT_LEAVE_FIELD_REDIRECT` can now send an Extra Deck monster (or a Pendulum) to the Extra Deck. Synchro Monsters Special Summoned by Angelechy Seneschal / 「具象天使的元帅」(`101306091`) return to the Extra Deck when they leave the field, instead of going to the GY.
- **ZH:** `EFFECT_LEAVE_FIELD_REDIRECT` 现在可以把额外卡组怪兽（或灵摆）送回额外卡组。具象天使的元帅（`101306091`）效果特殊召唤的同调怪兽离场时回到额外卡组，不再送去墓地。

## Background / 背景

`101306091` registers `EFFECT_LEAVE_FIELD_REDIRECT` with `SetValue(LOCATION_EXTRA)` on itself and on each Fairy Synchro it Special Summons. The effect is meant to replace the leave-field destination with the Extra Deck.

`101306091` 给自己，以及效果特殊召唤的每只天使族同调怪兽，注册 `EFFECT_LEAVE_FIELD_REDIRECT`，值为 `LOCATION_EXTRA`。离场时应用这条重定向，目的地改为额外卡组。

```lua
e3:SetCode(EFFECT_LEAVE_FIELD_REDIRECT)
e3:SetValue(LOCATION_EXTRA)
sc:RegisterEffect(e3,true)
```

## Problem / 问题

`card::leave_field_redirect` only kept `LOCATION_HAND`, `LOCATION_DECK`, and `LOCATION_REMOVED`. `LOCATION_EXTRA` (`0x40`) matched none of those branches, so the function returned `0`. `field::send_to` only replaces the destination when that value is non-zero, and the monster stayed on the GY path.

`card::leave_field_redirect` 原先只接受手牌、卡组、除外。`LOCATION_EXTRA`（`0x40`）三条都不进，函数返回 `0`。`field::send_to` 只在返回值非 0 时改写目的地，怪兽仍按送墓处理。

`RESETS_REDIRECT`, `EFFECT_FLAG_CANNOT_DISABLE`, and `RegisterEffect(..., true)` were already correct. The effect was read, then dropped because the destination was unsupported.

`RESETS_REDIRECT`、`EFFECT_FLAG_CANNOT_DISABLE` 和 `RegisterEffect(..., true)` 本来就没问题。效果能被读到，只是目的地不被接受。

## Fix / 修正

Core only. The card script is unchanged. Rebuild `ocgcore` before the change applies in a duel.

只改 core，卡片脚本不动。重新编译 `ocgcore` 后，对局里才会生效。

### `card::leave_field_redirect`

Accept `LOCATION_EXTRA` when `is_capable_send_to_extra` is true: Extra Deck monsters, or Pendulums, and not blocked by "cannot return to the Deck". Tokens still return `0` before this loop.

在 `is_capable_send_to_extra` 成立时接受 `LOCATION_EXTRA`：额外卡组怪兽或灵摆，且没有被「不能回卡组」挡住。衍生物仍在循环前直接返回 `0`。

If several leave-field redirects apply at once, priority is banish, then Deck, then Extra, then hand.

同时存在多种离场去向时，顺序是除外、卡组、额外、手牌。

### `field::send_to`

When the redirect changes the destination **to** the Extra Deck, set the position before `move_card`:

重定向把目的地**改成**额外卡组时，在 `move_card` 之前写表示形式：

- Fusion / Synchro / Xyz / Link: `POS_FACEDOWN_DEFENSE`.
- A Pendulum whose previous destination was the GY: `POS_FACEUP_DEFENSE`, same constant as the existing Pendulum rule in `field::move_card`.

- 融合 / 同调 / 超量 / 连接：`POS_FACEDOWN_DEFENSE`。
- 原本要送去墓地的灵摆：`POS_FACEUP_DEFENSE`，和 `field::move_card` 里原有灵摆规则用同一个常量。

Extra Deck placement only tests `position & POS_FACEUP`. `POS_FACEUP_DEFENSE` (`0x4`) satisfies that, so the Pendulum joins the face-up Extra pile and `extra_p_count` increases. Defense is not a separate Extra Deck rule; it keeps this path on the same position the old GY-to-Extra Pendulum path already writes. The code comment marks the Pendulum redirect ruling as still needing confirmation.

额外卡组入区只判断 `position & POS_FACEUP`。`POS_FACEUP_DEFENSE`（`0x4`）满足这个判断，灵摆进入表侧额外并增加 `extra_p_count`。守备不是额外卡组另有的规则，只是让这条路径和原来「送墓改道回额外」写下的表示形式相同。代码注释标明灵摆走这条重定向的裁定仍待确认。

If the destination was already the Extra Deck (`Duel.SendtoExtra`), the position is left alone.

目的地本来就是额外卡组（`Duel.SendtoExtra`）时，不改表示形式。

### `card::is_capable_cost_to_extra`

A cost that returns the card to the Extra Deck stays payable when leave-field redirect resolves to `LOCATION_EXTRA`. Previously that function required the final destination to be `LOCATION_DECK`, so this redirect would have made the cost illegal.

回到额外卡组的代价，在离场重定向结果是 `LOCATION_EXTRA` 时仍然可以支付。原先要求最终目的地必须是 `LOCATION_DECK`，这条重定向会把代价判成不能支付。


Wind2009-Louse:
Try to redirect to deck instead of extra deck

